PR #15, opened 2022-04-08
Print Eint instead of Edisp

dmejiar:
`edisp` was incorrectly passed as argument when printing the "interaction energy"


awvwgk (approved):
Nice catch. Thanks for sharing,

PR #52, opened 2022-05-30
Adds "angular-momenta" to Calculator class _getter dictionary.

kvkarandashev:
Created a function for exporting angular momenta of calculator's atomic orbitals, as discussed in #49

### DIFF
--- a/include/tblite/calculator.h
+++ b/include/tblite/calculator.h
@@ -113,6 +113,12 @@ tblite_get_calculator_shell_map(tblite_context /* ctx */,
                             tblite_calculator /* calc */,
                             int* /* sh2at */);
 
+/// Query calculator for angular momenta of shells
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_get_calculator_shell_angular_momenta(tblite_context /* ctx */,
+                              tblite_calculator /* calc */,
+                              int* /* sh_am */);
+
 /// Query calculator for the number of orbitals
 TBLITE_API_ENTRY void TBLITE_API_CALL
 tblite_get_calculator_orbital_count(tblite_context /* ctx */,
@@ -124,13 +130,6 @@ TBLITE_API_ENTRY void TBLITE_API_CALL
 tblite_get_calculator_orbital_map(tblite_context /* ctx */,
                               tblite_calculator /* calc */,
                               int* /* ao2sh */);
-
-/// Query calculator for angular momenta of atomic orbitals
-TBLITE_API_ENTRY void TBLITE_API_CALL
-tblite_get_calculator_ao_angular_momenta(tblite_context /* ctx */,
-                              tblite_calculator /* calc */,
-                              int* /* ao_am */);
-
 
 /// Perform single point calculation
 TBLITE_API_ENTRY tblite_calculator TBLITE_API_CALL
--- a/include/tblite/calculator.h
+++ b/include/tblite/calculator.h
@@ -116,8 +116,8 @@ tblite_get_calculator_shell_map(tblite_context /* ctx */,
 /// Query calculator for angular momenta of shells
 TBLITE_API_ENTRY void TBLITE_API_CALL
 tblite_get_calculator_angular_momenta(tblite_context /* ctx */,
-                              tblite_calculator /* calc */,
-                              int* /* am */);
+                                      tblite_calculator /* calc */,
+                                      int* /* am */);
 
 /// Query calculator for the number of orbitals
 TBLITE_API_ENTRY void TBLITE_API_CALL
--- a/include/tblite/calculator.h
+++ b/include/tblite/calculator.h
@@ -125,6 +125,13 @@ tblite_get_calculator_orbital_map(tblite_context /* ctx */,
                               tblite_calculator /* calc */,
                               int* /* ao2sh */);
 
+/// Query calculator for angular momenta of atomic orbitals
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_get_calculator_ao_angular_momenta(tblite_context /* ctx */,
+                              tblite_calculator /* calc */,
+                              int* /* ao_am */);
+
+
 /// Perform single point calculation
 TBLITE_API_ENTRY tblite_calculator TBLITE_API_CALL
 tblite_get_singlepoint(tblite_context /* ctx */,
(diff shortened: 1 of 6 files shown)


awvwgk (changes requested):
Preferably, we can return the shell angular momenta rather than the orbital ones, using the `orbital-map` those can be easily expanded and don't require to produce redundant information in the API.

Also, please add tests for all your changes.

kvkarandashev:
Thank you for the feedback! I decided to make API's new output part of the test_gfn2 procedure in test_interface, but I am afraid I have trouble passing this test both with my branch and tblite/tblite main branch. On closer inspection, for tblite/tblite main branch I can pass all tests in tblite.test_interface but for test_gfn2 and test_gfn2_mindless; I also checked that running "meson test -C _build --print-errorlogs" passes all tests but for "xtb-param" and "xtb-external", for which it faces a timeout.
Could you check whether the issue is with the tests or with my installation?

awvwgk:
>  Could you check whether the issue is with the tests or with my installation?

Depending on your machine setup, it might be worth to limit the number of threads and/or concurrent runs to avoid over-subscription of the physical cores:

```sh
export OMP_NUM_THREADS=$(nproc),1
meson test -C _build --print-errorlogs --num-processes 1
```

kvkarandashev:
Thank you! Running "meson test" with only one process and one OMP thread did solve the timeout issue; I now pass 33 tests with 1 expected failure ("tblite:app / noargs").

However, I am afraid that the problem with reproducing "test_interface.test_gfn2" and "test_interface.test_gfn2_mindless" persists as I fail "assert res.get("energy") == ..." in both cases despite passing "tblite:unit / gfn2-xtb" of the meson test.

awvwgk:
> However, I am afraid that the problem with reproducing "test_interface.test_gfn2" and "test_interface.test_gfn2_mindless" persists as I fail "assert res.get("energy") == ..." in both cases despite passing "tblite:unit / gfn2-xtb" of the meson test.

Strange indeed, the setup for the continuous integration passes the Python API with your modification, so it should be fine I guess.

awvwgk:
Would `angular_momenta` be sufficient here to keep the name short?

kvkarandashev:
I have changed python API to return angular momenta corresponding to shells rather than AO's and added the corresponding test to test_gfn2 in tblite.test_interface. Unfortunately, I still have trouble passing this test due to energy assertion error, but the latter is raised after the angular momenta value assertion is passed successfully.
I will revisit the tests in test_interface sometime this week and if the problem is still there and I reproduce it on a different machine I will create an issue for it.

kvkarandashev:
> Would `angular_momenta` be sufficient here to keep the name short?

Now that I think about it yes, it is more fitting than "shell_angular_momenta". I have made the corresponding commit.

include/tblite/calculator.h
```diff
@@ -113,6 +113,12 @@ tblite_get_calculator_shell_map(tblite_context /* ctx */,
                             tblite_calculator /* calc */,
                             int* /* sh2at */);
 
+/// Query calculator for angular momenta of shells
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_get_calculator_angular_momenta(tblite_context /* ctx */,
+                              tblite_calculator /* calc */,
+                              int* /* am */);
```
awvwgk: Hanging indentation for C:

```suggestion
tblite_get_calculator_angular_momenta(tblite_context /* ctx */,
                                      tblite_calculator /* calc */,
                                      int* /* am */);
```

awvwgk (changes requested):
Please also add tests for the C-API in [`test/api/main.c`](https://github.com/tblite/tblite/blob/main/test/api/main.c).

Especially the failing case at

https://github.com/tblite/tblite/blob/4a82f966e919311e7cb489f26e298df5592e6bc5/test/api/main.c#L254

src/tblite/api/calculator.f90
```diff
@@ -401,6 +401,35 @@ subroutine get_calculator_shell_map(vctx, vcalc, imap) &
...
 
+subroutine get_calculator_angular_momenta(vctx, vcalc, am) &
+      & bind(C, name=namespace//"get_calculator_angular_momenta")
+   type(c_ptr), value :: vctx
+   type(vp_context), pointer :: ctx
+   type(c_ptr), value :: vcalc
+   type(vp_calculator), pointer :: calc
+   integer(c_int), intent(inout) :: am(*)
+   type(error_type), allocatable :: error
+   integer:: ish
```
awvwgk:
```suggestion
   integer :: ish
```

src/tblite/api/calculator.f90
```diff
@@ -401,6 +401,35 @@ subroutine get_calculator_shell_map(vctx, vcalc, imap) &
...
+      call fatal_error(error, "Calculator object is missing")
+      call ctx%ptr%set_error(error)
+      return
+   end if
+   call c_f_pointer(vcalc, calc)
+
+   associate(bas => calc%ptr%bas)
+      do ish=1, bas%nsh
+         am(ish)=(bas%nao_sh(ish)-1)/2
+      enddo
```
awvwgk: That's a clever way to retrieve the angular momentum of a shell, clever implementations deserve a comment explaining why they were done this way to spare future generations the mental exercise to figure out what is going on here.
```suggestion
      do ish = 1, bas%nsh
         ! The angular momentum is only available for each species,
         ! however the species to atom mapping is unavailable in this API call.
         ! Therefore, we calculate it by using 2*l + 1 == nao
         am(ish) = (bas%nao_sh(ish)-1)/2
      end do
```

awvwgk (commented):
Some final style comments

test/api/main.c
```diff
@@ -971,6 +987,31 @@ test_gfn2_si5h12 (void)
...
+
+   tblite_get_calculator_shell_map(ctx, calc, sh2at);
+   if (tblite_check_context(ctx)) goto err;
+
+   if (!check(sum_int(nshells, sh2at), 156, "dimension error")) goto err;
+
+   tblite_get_calculator_angular_momenta(ctx, calc, am);
+   if (tblite_check_context(ctx)) goto err;
+
+   if (!check(sum_int(nshells, am), 15, "dimension error")) goto err;
```
awvwgk: Summing the angular momenta doesn't make sense, better check for a specific atom, like a hydrogen where the momenta are clear. Also the error message is incorrect in this context.

kvkarandashev:
OK, would it be even better to compare the integer array to a benchmark array entry-wise, just like in test_interface.test_gfn2? Or would it be too much data stored for the test?

awvwgk (approved):
Thanks for sharing, looks good to me now.

kvkarandashev:
Thanks a lot!

PR #55, opened 2022-06-02
Add `--iterations <int>` as  option

marvinfriede:
- allow user to set maximum number of SCF iterations
- naming in accordance with xTB (`--iterations`)


awvwgk (approved):
Thanks for sharing. This is a useful addition.

PR #87, opened 2022-10-19
Add verbosity option to ASE Calculator

zyt0y:
Current ASE-TBlite Calculator does not recognize `verbosity` option. I have enabled this option in ASE TBlite Calculator.


awvwgk (changes requested):
Please add the verbosity to the `default_parameters` dict as well.

zyt0y:
`verbosity` is now added to doc and `default_parameters` dict

awvwgk (approved):
Thanks for sharing, looks good to me now.

PR #90, opened 2022-11-16
Fixing the spGFN gradient

haneug:
Maybe a little primitive. I am open to suggestions.


haneug:
I see I will have to update the test suite as well.

awvwgk (commented):
I made a proposal to avoid the if condition in the spin loop. Please have a close look whether that works as expected.

haneug:
Nice proposal! I had to make the spin iterator explicit after the do loop, else 'spin' will be larger 'nspin'. The results remain the same as with my initial commit.

awvwgk (approved):
Thanks for sharing looks good to me.

PR #92, opened 2022-11-30
Hamiltonian gradient: Avoid some repeated computations

marvinfriede:
- store intermediate results in computation of shell polynomials for gradient
- move shell polynomial part outside AO loop

These are only some minor improvements, nothing too important. Just wanted to raise the issue, while I was going through the code anyway.


awvwgk (approved):
Thanks for sharing. Looks good to me.

PR #118, opened 2023-06-21
implementation of xtbml features

pitsteinbach:
Hey,

I implemented our so called xtbml features in tblite.
Therefore, I added two CLI arguments `--xtbml` and `--xtbml_xyz`, the difference being that the latter exports the features in their tensorial form and not only the norm.

The features are written into a csv file when using CLI, the relevant getters and setters are also implemented to compute the features via the python API. Here the part where I would be looking for some help is to transfer the feature labels as an array to python. Currently this is done using repeated API calls with varying index values, so not elegant neither efficient.

I needed to implement some specific subroutines for the anisotropic energies to match our previous xtb based values. 

I added a few unit tests mostly relevant to the CAMMs and their extensions but also checking some invariances. Validation tests that are comparing against known features for charged open shell molecules are used, as well as checking the tensorial output gainst benzene.

Of course I am open for suggestions.

Greets
Pit


src/tblite/wavefunction/type.f90
```diff
@@ -146,5 +154,67 @@ subroutine get_alpha_beta_occupation(nocc, nuhf, nalp, nbet)
    nbet = ntmp / 2
 end subroutine get_alpha_beta_occupation
 
+subroutine occu(ndim,nel,nopen,ihomoa,ihomob,focca,foccb)
```
awvwgk: This is an artifact from the xtb code base, why do we need to introduce it here if we already have procedures for handling this case?

src/tblite/xtb-ml/atomic_frontier_orbitals.f90
```diff
@@ -0,0 +1,394 @@
...
+          egap(m)=1.0_wp/(egap(m)+1.0d-12)-damp
+          ehoaoa(m)=chempot(m)
+          eluaoa(m)=chempot(m)+egap(m)
+        enddo !  occ
+      endif
+   enddo
+    
+   if (print) then
+      do  m=1,nat
+         write(*,'(3x,i6,5x,f8.4,5x,f8.4,5x,f8.4,5x,f8.4,5x,f8.4)') m,response(m)*(autoev**2),egap(m),chempot(m),ehoaoa(m),eluaoa(m)
```
awvwgk: Routines in tblite must not call intrinsic print but use the logger.

src/tblite/xtb-ml/xtbml.f90
```diff
@@ -0,0 +1,185 @@
...
+! tblite is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @file tblite/xtb-ml/xtbml.f90
+module xtbml_base
```
awvwgk: Please prefix your modules with `tblite_` (see https://fpm.fortran-lang.org/spec/naming.html)

awvwgk (changes requested):
Thanks for your contribution. I will put some time aside for reviewing, however before I start I would kindly ask you to follow the style guide for the respective files:

- Python: black
- C: hanging indentation
- Fortran: indent function bodies, whitespace around operators, commas, assignments, ..., no trailing whitespace

src/tblite/wavefunction/type.f90
pitsteinbach: I removed the function and got the quantity from other variables

awvwgk:
The CMake build is currently broken:

````
[759/804] Building Fortran object CMakeFiles/tblite-lib.dir/src/tblite/xtb/singlepoint.f90.o
FAILED: CMakeFiles/tblite-lib.dir/src/tblite/xtb/singlepoint.f90.o include/tblite_xtb_singlepoint.mod 
/usr/bin/gfortran -I/home/runner/work/tblite/tblite/src/tblite/xtb -I/home/runner/work/tblite/tblite/include -I/home/runner/work/tblite/tblite/_build/include -I/home/runner/work/tblite/tblite/_build/_deps/toml-f-build/include -I/home/runner/work/tblite/tblite/_build/_deps/mctc-lib-src/include -I/home/runner/work/tblite/tblite/_build/_deps/mctc-lib-build/include -I/home/runner/work/tblite/tblite/_build/_deps/dftd4-src/include -I/home/runner/work/tblite/tblite/_build/_deps/dftd4-build/include -I/home/runner/work/tblite/tblite/_build/_deps/multicharge-build/include -I/home/runner/work/tblite/tblite/_build/_deps/s-dftd3-src/include -I/home/runner/work/tblite/tblite/_build/_deps/s-dftd3-build/include -g -Jinclude -fPIC -fopenmp -fpreprocessed -c CMakeFiles/tblite-lib.dir/src/tblite/xtb/singlepoint.f90-pp.f90 -o CMakeFiles/tblite-lib.dir/src/tblite/xtb/singlepoint.f90.o
/home/runner/work/tblite/tblite/src/tblite/xtb/singlepoint.f90:46:8:

   46 |    use xtbml_base, only : xtbml_base_type
      |        1
Fatal Error: Cannot open module file ‘xtbml_base.mod’ for reading at (1): No such file or directory
compilation terminated.
````

src/tblite/xtb-ml/atomic_frontier_orbitals.f90
pitsteinbach: using a call to context message instead

src/tblite/xtb-ml/xtbml.f90
pitsteinbach: changed

awvwgk:
Pit @pitsteinbach, do you need any help with incorporating the changes?

pitsteinbach:
Thanks for asking, in fact I am not really sure on how to proceed, are you referring to the warnings from CodeCov? I have not yet used this change request feature from GitHub, if you could hit me up with a short mail on how to move forward I would be happy to incorporate the changes myself.

pitsteinbach:
bump.

awvwgk:
The review comments are still unaddressed, let's organize a meeting this week to make some progress here.

pitsteinbach:
I will close this pull request and open several ones after segmenting the introduced feature.

I will separate the categories in types that can be attached to a xtbml class individually and also separate the computation of features and extended features. 
I will create a dictionary class to hold the key value pairs of each category and then concatenate them to get the dictionary.

PR #123, opened 2023-08-01
corrected minor error in the principal quantum number of Cs and Ba

marcelmbn:
- Correct wrong principal quantum numbers in GFN1-xTB for Cs and Ba (probably without consequences; was presumably taken over from the SI of the GFN1-xTB paper, in which these values are also wrong)
- For the sake of readability, commented "Z ranges" for each row are added for the three 7x13 matrices (probably a matter of taste)


awvwgk:
Can you please verify whether these values are correct compared to `xtb` or whether a fix is needed there as well?

- parameter file: https://github.com/grimme-lab/xtb/blob/8d0f1dd02f9d00324d95662d90f3fa123bf01e1f/param_gfn1-xtb.txt#L746-L770
- source code: https://github.com/grimme-lab/xtb/blob/8d0f1dd02f9d00324d95662d90f3fa123bf01e1f/src/xtb/gfn1.f90#L255-L269

While a principal quantum number of 6 rather than 5 seems more reasonable here, this seems to be a long standing bug in `xtb`.

marcelmbn:
> Can you please verify whether these values are correct compared to `xtb` or whether a fix is needed there as well?
> 
>     * parameter file: https://github.com/grimme-lab/xtb/blob/8d0f1dd02f9d00324d95662d90f3fa123bf01e1f/param_gfn1-xtb.txt#L746-L770
> 
>     * source code: https://github.com/grimme-lab/xtb/blob/8d0f1dd02f9d00324d95662d90f3fa123bf01e1f/src/xtb/gfn1.f90#L255-L269
> 

In my understanding, both of the files should be changed accordingly. As mentioned in the original post of this PR, this mistake is already present in the SI of the original paper. Probably, this error was then never changed over time.
If you agree on this, I'd open a PR for these files as well.

> 
> While a principal quantum number of 6 rather than 5 seems more reasonable here, this seems to be a long standing bug in `xtb`.

I agree. Am I right with the assumption that this, however, does not lead to any practical consequences for calculations including Cs/Ba?

awvwgk:
Changing the principal quantum number changes the basis set and impacts the parametrization of GFN1-xTB. I recall the last time we did this for sodium and magnesium (around 2017) due to wrong quantum numbers (2s instead of 3s) the parameters were reoptimized accordingly. The parameter file in the `xtb` repository is mostly unmodifed from the original version and the SI of the paper reflects this correctly. I agree that the principal quantum number is incorrect, however just changing it in the parameter file will be incorrect as well, without doing proper verification for the impacted elements.

PR #124, opened 2023-08-26
Implementation of Charge-Extended Hückel method

marcelmbn:
### Full implementation of Charge-Extended Hückel (CEH) method as guess and charge model in tblite

Theoretical foundation can be obtained upon request or (shortly) in "Müller, M., Hansen, A. & Grimme, S. An atom-in-molecule adaptive polarized valence single-ζ atomic orbital basis for electronic structure calculations. _submitted to J. Chem. Phys._". The publication also gives MAEs for atomic charges and dipole moments.
CEH serves as a charge model for:
1. An atomic charge guess for the new GP3-xTB tight binding method
2. The atom-in-molecule-adaptive q-vSZP basis set as stand-alone
3. An adapted variant of q-vSZP utilized in the GP3-xTB TB method

The implementation contains (besides the CEH-specific Hamiltonian)
- An electronegativity difference-weighted coordination number
- Overlap integrals that are scaled in the diatomic frame depending on the diatomic interaction type (_sigma_, _pi_, _delta_)

CEH is by purpose not implemented as another tight-binding Hamiltonian parallel to GFN1/2 but as a guess method (along with EEQ/SAD) providing atomic charges (or a complete density matrix). The main reasons are that (i) CEH is not a tight-binding method but "only" an EHT-based model, and (ii) it provides only a density matrix but has neither an energy expression nor analytical gradients available.
The implementation uses (wherever possible and reasonable) the existing tblite infrastructure (e.g., for integrals, electronic solvers, wavefunction, ...) and includes unit tests assessing charges and dipole moments of charged molecules, probes the electric field dependency, the scaled overlap contributions and the EN-weighted coordination number.


marcelmbn:
The CEH feature (not using it as a guess but as a "stand-alone" result) is accessible via the command-line call:
`tblite run --guess ceh --guessonly coord`

src/tblite/ceh/ncoord_ceh/covrad_ceh.f90
```diff
@@ -0,0 +1,94 @@
...
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @file tblite/ceh/ncoord_ceh/covrad_ceh.f90
+!> Provides covalent radii for all elements (up to 86)
+!> Covalent radii were fitted together with other empirical parameters # MM, August 02, 2023
+
+module tblite_data_covrad_ceh
```
awvwgk: Please make sure that the module name can be mapped to the file name. This file should be in `src/tblite/data`

src/tblite/ceh/ncoord_ceh/covrad_ceh.f90
```diff
@@ -0,0 +1,94 @@
...
+   use mctc_env, only : wp
+   use mctc_io_convert, only : aatoau
+   use mctc_io_symbols, only : to_number
+   implicit none
+   private
+
+   public :: get_covalent_rad
```
awvwgk: This symbol name can overlap with existing symbols, maybe rename it?

src/tblite/ceh/ncoord_ceh/ncoord_ceh.f90
```diff
@@ -0,0 +1,57 @@
...
+!
+! tblite is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @dir tblite/ncoord
```
awvwgk: This direction instruction does not match with the file name

src/tblite/ceh/ncoord_ceh/ncoord_ceh_standard.f90
```diff
@@ -0,0 +1,336 @@
...
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @file tblite/ceh/ncoord_ceh/ncoord_ceh_standard.f90
+!> Provides a coordination number implementation for the CEH method
+
+!> Coordination number implementation for the CEH method with two different coordination numbers
+!> EN-weighted and usual form.
```
awvwgk: Can those two be merged in a single interface as done for dftd4?

src/tblite/ceh/ncoord_ceh/ncoord_ceh_standard.f90
```diff
@@ -0,0 +1,336 @@
...
+                  cn_en(jat) = cn_en(jat) + countf * ( en(mol%num(izp)) - en(mol%num((jzp))) )
+               end if
+            end do
+         end do
+      end do
+
+   end subroutine ncoord_erf
+
+
+! subroutine ncoord_derf(mol, trans, cutoff, rcov, en, cn, dcndr, dcndL)
```
awvwgk: Please remove unused code

src/tblite/ceh/ncoord_ceh/type_ceh.f90
```diff
@@ -0,0 +1,51 @@
...
+
+!> Declaration of base class for coordination number evalulations
+module tblite_ncoord_type_ceh
+   use mctc_env, only : wp
+   use mctc_io, only : structure_type
+   implicit none
+   private
+
+   !> Abstract base class for coordination number evaluator
+   type, public, abstract :: ncoord_type_ceh
```
awvwgk: Why is a separate base class needed?

awvwgk (changes requested):

src/tblite/ceh/ncoord_ceh/ncoord_ceh_standard.f90
marcelmbn: Done.

src/tblite/ceh/ncoord_ceh/type_ceh.f90
marcelmbn: Refactored the setup of the coordination number such that it is included in the `tblite_ncoord` framework and not as a stand-alone directory with own classes/types.

src/tblite/ceh/ncoord_ceh/ncoord_ceh_standard.f90
marcelmbn: See below.

src/tblite/ceh/ncoord_ceh/ncoord_ceh.f90
marcelmbn: Obsolete now.

src/tblite/ceh/ncoord_ceh/covrad_ceh.f90
marcelmbn: Done.

src/tblite/ceh/ncoord_ceh/covrad_ceh.f90
marcelmbn: Done.

src/tblite/wavefunction/type.f90
```diff
@@ -66,6 +66,45 @@ module tblite_wavefunction_type
       real(wp), allocatable :: qpat(:, :, :)
    end type wavefunction_type
 
+   !> Tight-binding wavefunction derivative
+   type, public :: wavefunction_derivative_type
```
awvwgk: Why does this need its own type, can't it be incorporated with the normal wavefunction type and allocated on demand when the wavefunction is constructed?

test/unit/test_integral_multipole.f90
```diff
@@ -43,6 +43,7 @@ subroutine collect_integral_multipole(testsuite)
    type(unittest_type), allocatable, intent(out) :: testsuite(:)
 
    testsuite = [ &
+      new_unittest("overlap_dipole_diat_alh3", test_overlap_dipole_diat_alh3), &
```
awvwgk: Please prefer dashes in test names

app/cli.f90
```diff
@@ -291,6 +293,9 @@ subroutine get_run_arguments(config, list, start, error)
             call fatal_error(error, "Missing argument for method")
             exit
          end if
+      
+      case("--guessonly")
+         config%guessonly = .true.
```
awvwgk: I would prefer to have this either as a standalone (`--method ceh`) or have a separate subcommand `tblite guess --method {sad,ceh,...}`. The logic of `--guessonly` is not fitting for the run driver at the moment.

src/tblite/wavefunction/type.f90
marcelmbn: The wavefunction derivative type was implemented as a starting point for future implementation of the CEH gradient. But since I haven't thought about if it will actually be required, I deleted it and all references for the moment and cleaned up everything. 
This (or your suggestion) can be implemented consistently in future gradient development when it is actually required.

test/unit/test_integral_multipole.f90
marcelmbn: Done.

app/cli.f90
marcelmbn: (I haven't completed the change towards a new subcommand 'guess' yet, so it is to be expected that it will not work yet.)
I'll request a review when it is done.

marcelmbn: Done.

app/driver_run.f90
```diff
@@ -145,24 +151,45 @@ subroutine run_main(config, error)
...
+
    select case(config%guess)
    case default
       call fatal_error(error, "Unknown starting guess '"//config%guess//"' requested")
    case("sad")
       call sad_guess(mol, calc, wfn)
    case("eeq")
       call eeq_guess(mol, calc, wfn)
+   case("ceh")
+      ctx%verbosity = config%verbosity
```
awvwgk: The guess must not modify the verbosity of the calculation context.

app/driver_run.f90
```diff
@@ -145,24 +151,45 @@ subroutine run_main(config, error)
...
+      call ceh_guess(ctx, calc_ceh, mol, error, wfn_ceh)
```
awvwgk: Since we pass in the calculation context here, we should check whether an error was recorded in the context and needs handling.

src/tblite/ceh/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+set(dir "${CMAKE_CURRENT_SOURCE_DIR}")
+
+list(
+  APPEND srcs
+  "${dir}/ceh.f90"
+  "${dir}/ceh_h0.f90"
+  "${dir}/ceh_calculator.f90"
```
awvwgk: The prefix `ceh_` is superfluous here.

app/driver_guess.f90
```diff
@@ -68,7 +68,6 @@ subroutine guess_main(config, error)
 
       ctx%terminal = context_terminal(config%color)
       ctx%solver = lapack_solver(config%solver)
-      ctx%verbosity = config%verbosity
```
awvwgk: Setting the verbosity globally via the context here and in run is okay, but having it dependent on the guess is not.

marcelmbn: Okay, got it. 
In the case of `guess_main` it doesn't really matter in the end in my eyes (if it is set globally in the beginning or if the `config%verbosity` is taken). I just decided to proceed analogously to `run_main` (or how it was done before), in which `xtb_singlepoint` takes `config%verbosity` as an argument. In `guess_main`, the analog routine is the call of `ceh_guess`.

marcelmbn:
Are further steps required for the final integration into `main` of `tblite`?

awvwgk (changes requested):
There is one thing that is currently unclear to me. When doing the Slater-Koster rotation of the integral in the diatomic frame for scaling, the transformation is done with the whole matrix rather than the contribution of the pair that is rotated. This step seems incorrect since it will destroy the integral matrix if multiple images of the atom are present. Rather the transformation needs to happen on the diatomic block level.

src/tblite/integral/overlap.f90
```diff
@@ -452,5 +454,132 @@ subroutine get_overlap_lat(mol, trans, cutoff, bas, overlap)
...
+      integer :: nao_ati, nao_atj, lbi, lbj, ubi, ubj
+      integer :: ish, jsh, ii, jj, iao, jao, nao, maxl_ish_jsh
+      real(wp) :: r2, vec(3), cutoff2, ksig, kpi, kdel
+      real(wp), allocatable :: stmp(:)
+      real(wp) :: block_overlap(9,9)
+
+      if (size(scal_fac,1) /= 3) then
+         write(*,*) 'Error: scal_fac must have the dimension of 3, &
+         & since it covers the three different types of bonding'
+         stop
```
awvwgk: This should be an error stop. Also, please don't write a separate message and than stop, rather directly write with the stop statement.

src/tblite/ceh/h0.f90
```diff
@@ -0,0 +1,33 @@
...
+!
+! tblite is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @file tblite/ceh/ceh_h0.f90
```
awvwgk:
```suggestion
!> @file tblite/ceh/h0.f90
```

src/tblite/ceh/calculator.f90
```diff
@@ -0,0 +1,127 @@
...
+!
+! tblite is distributed in the hope that it will be useful,
+! but WITHOUT ANY WARRANTY; without even the implied warranty of
+! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+! GNU Lesser General Public License for more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+!> @file tblite/ceh/ceh_calculator.f90
```
awvwgk:
```suggestion
!> @file tblite/ceh/calculator.f90
```

src/tblite/ceh/ceh.f90
```diff
@@ -0,0 +1,777 @@
...
+      &  overlap, overlap_diat, dipole)
+
+      !> define off-diagonal elements of CEH Hamiltonian
+      k = 0
+      do iat = 1, mol%nat
+         offset_iat = calc%bas%ish_at(iat)
+         do ish = 1, calc%bas%nsh_at(iat)
+            do iao = 1, calc%bas%nao_sh(ish + offset_iat)
+               !> iterator for AO of i increased by 1 -> k
+               k = k + 1
```
awvwgk: Please compute those indices directly from your basis set rather than manually incrementing them from the outer loop. This avoids having unreachable error stops in this code block.

awvwgk (changes requested):
There are some minor formatting issues where the code is currently not following the style of the rest of tblite, please check and reformat the files accordingly.

marcelmbn:
> There are some minor formatting issues where the code is currently not following the style of the rest of tblite, please check and reformat the files accordingly.

Which parts/formatting issues do you mean, for example? How can I reformat them so that they match the rest of the `tblite` style? (Do you mean some `findent`/`fprettify` settings?)

src/tblite/integral/overlap.f90
```diff
@@ -452,5 +454,132 @@ subroutine get_overlap_lat(mol, trans, cutoff, bas, overlap)
...
+                  ! trans_block_s = O^T * S * O
+                  call diat_trafo(block_overlap, vec_diat_trafo, ksig, kpi, kdel)
+                  ! (see 'optional (1.1)' above)
+                  ! trans_block_s = transpose(trans_block_s)
+                  !> 4. Fill the overlap_diat matrix with the back-transformed submatrix
+                  overlap_scaled(lbj:ubj, lbi:ubi) = block_overlap(1:nao_atj, 1:nao_ati)
+               else
+                  !> 5. Fill the overlap_scaled matrix with the plain overlap submatrix (no transformation)
+                  overlap_scaled(lbj:ubj, lbi:ubi) = overlap(lbj:ubj, lbi:ubi)
+               endif
```
marcelmbn: > There is one thing that is currently unclear to me. When doing the Slater-Koster rotation of the integral in the diatomic frame for scaling, the transformation is done with the whole matrix rather than the contribution of the pair that is rotated. This step seems incorrect since it will destroy the integral matrix if multiple images of the atom are present. Rather the transformation needs to happen on the diatomic block level.

In the above lines of code, only the pair contribution (just like you would have expected) is rotated into the diatomic frame. I.e., first, the overlap between _all_ AOs in _all_ shells between atoms `iat` and `jat` is calculated and set up. Then, this block matrix in the overlap (lines 544 to 554) is put into the transformation routine (transform forth, scale, transform back), and is added onto the second overlap matrix `overlap_scaled`. Does this answer your question?

src/tblite/integral/overlap.f90
marcelmbn: Done.

src/tblite/ceh/ceh.f90
marcelmbn: Done.

marcelmbn:
> There is one thing that is currently unclear to me. When doing the Slater-Koster rotation of the integral in the diatomic frame for scaling, the transformation is done with the whole matrix rather than the contribution of the pair that is rotated. This step seems incorrect since it will destroy the integral matrix if multiple images of the atom are present. Rather the transformation needs to happen on the diatomic block level.

Based on your comment and our in-person discussion, I refactored the rotation of the overlap integral elements such that it is performed on the shell level. I.e., it can used when calculating only the overlap integral elements of a single shell pair, as it is regularly done for, e.g., the setup of the GFN1/2 Hamiltonian.
Consequently, I adapted the setup of the CEH Hamiltonian to this working principle. Instead of calling a subroutine for the whole overlap matrix at the beginning of the Hamiltonian setup, the overlap integral is built step-by-step, avoiding unnecessary calculations for the symmetry-equivalent triangle.

awvwgk (approved):
Looks good to me now

PR #125, opened 2023-09-06
Add double dictionary

pitsteinbach:
I added a dictionary derived type that takes doubles arrays as entries, the size of the array can vary by entry.
The dimension of the array is currently explicitly coded for 3 dimensions, maybe a smarter solution is possible.
A lookup is possible via the label of the entry or its index, returning the array if the appropriate size of array is given otherwise the array stays unallocated same if the label or index is not in the dictionary.
The labels can also be looked up via the index, which allows to print out or pack the content of the dictionary in an easy fashion.

I also added an explicit assignment operator as well as + operator to allow easy concatenation of dictionaries.

The introduction of this class will allow a more efficient collection of various ML feature categories, that will be subject to a future pull request. 
A testsuite for the functionality is also provided.

Thanks in advance
Pit

### DIFF
--- a/src/tblite/dictionary.f90
+++ b/src/tblite/dictionary.f90
@@ -49,10 +49,47 @@ module tblite_double_dictionary
       procedure :: copy
       generic :: operator(+) => combine_dict
       procedure :: combine_dict
+      generic, public :: remove_entry => remove_entry_label, remove_entry_index
+      procedure :: remove_entry_label
+      procedure :: remove_entry_index
    end type double_dictionary_type
 
 
 contains
+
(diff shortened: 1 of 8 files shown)


src/tblite/dictionary.f90
```diff
@@ -0,0 +1,490 @@
+module tblite_double_dictionary
```
awvwgk: The file should be named `src/tblite/double_dictionary.f90` also the license header is missing here as well as the Doxygen comments to include it in the generated documentation. Other than that it ready to get merged.

awvwgk (approved):
Looks good to me

PR #127, opened 2023-09-28
add bind(C) to callback function of context 

pitsteinbach:
I ran into issues, when compiling the project with newer intel compilers. The main issue I encountered was that when running singlepoint calculations from python, I get errors from cffi that the length of the char pointer is returned as negative. Therefore the passed message can not be decoded.

I managed to solve the issue by adding the bind(C) attribute to the callback function. Now everything works fine with intel oneAPI compilers (so far I tested with ifort 2021.10.0, more from our cluster following) as well.

### DIFF
--- a/src/tblite/api/context.f90
+++ b/src/tblite/api/context.f90
@@ -42,7 +42,7 @@ module tblite_api_context
 
    abstract interface
       !> Interface for callbacks used in custom logger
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Good catch, thanks for sharing

PR #128, opened 2023-10-10
Adding post processing container

pitsteinbach:
I added the construct of a post processing container to tblite. It's purpose is to collect post processing calculations and compute them after the converged singlepoint. It uses the double dictionary class to store the results inside.
For the first implementation , I added the Mayer-Wiberg bond calculation and molecular dipole and quadrupole moments as post processing calculations. Thereby removing there explicit calculation and rather putting them in the post processing container. I also corrected the bond orders for UHF wavefunctions.
To make use of the dictionary structure in other languages through the API, I added some calls to access the entries. In python the dictionary is reconstructed as a python dictionary. This could be especially beneficent for ML tasks. After this pull request for the general construct I am planning to add another pull request for the xtbml features.

I let the CI run once to check coverage by existing tests, and will then come back to adding tests.


include/tblite/calculator.h
```diff
@@ -210,3 +211,17 @@ tblite_get_singlepoint(tblite_context ctx,
...
+
+/// Perform single point calculation
+///
+/// @param ctx: Context handle
+/// @param mol: Molecular structure data
+/// @param calc: Calculator instance
+/// @param res: Result container
+/// @param post_process: Post processing container
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_get_singlepoint_w_post(tblite_context ctx,
```
awvwgk: Could this be added to the calculator handle and the dispatch handled on the Fortran side of the API?

pitsteinbach: good point, then we can use one singlepoint call.

pitsteinbach:
This feature would be ready now from my side. I also corrected some double dictionary tests, I forgot to check for error allocation in them but its corrected now.

pitsteinbach:
@awvwgk have you had a chance to review the rest of this PR yet?

app/cli.f90
```diff
@@ -382,6 +384,13 @@ subroutine get_run_arguments(config, list, start, error)
             exit
          end if
 
+      case("--post-processing")
+         iarg = iarg + 1 
+         call list%get(iarg, config%post_processing)
+         if (.not.allocated(config%post_processing)) then
+            call fatal_error(error, "Missing argument for post processing")
+            exit
+         end if
```
awvwgk, line 393:
```suggestion
         end if

```

app/driver_run.f90
```diff
@@ -220,13 +232,34 @@ subroutine run_main(config, error)
...
-
-   call xtb_singlepoint(ctx, mol, calc, wfn, config%accuracy, energy, gradient, sigma, &
-      & config%verbosity, results)
+   if (allocated(post_proc)) then
+      call xtb_singlepoint(ctx, mol, calc, wfn, config%accuracy, energy, gradient, sigma, &
+         & config%verbosity, results, post_proc)
+   else
+      call xtb_singlepoint(ctx, mol, calc, wfn, config%accuracy, energy, gradient, sigma, &
+         & config%verbosity, results)
+   end if
```
awvwgk:
```suggestion

   call xtb_singlepoint(ctx, mol, calc, wfn, config%accuracy, energy, gradient, sigma, &
      & config%verbosity, results, post_proc)
```

python/tblite/interface.py
```diff
@@ -540,12 +545,19 @@ def add(self, interaction, *args) -> None:
...
         """
 
-        if interaction not in self._interaction:
+        if interaction in self._interaction:
+            cont = self._interaction[interaction](self._ctx, self._mol, self._calc, *args)
+            library.calculator_push_back(self._ctx, self._calc, cont)
+        elif interaction in self._post_processing:
+            library.post_processing_push_back(self._ctx, self._calc, self._post_processing[interaction])
+        elif ".toml" in interaction:
+            library.post_processing_push_back(self._ctx, self._calc, self._post_processing[interaction])
```
awvwgk, line 554: This is not the right place to add this, I feel like we can have a much higher level API for these post processing tasks in Python rather than allowing to pass in arbitrary ones.

pitsteinbach: I agree, that this is not the most elegant solution. I don't really how you would like to achieve this, can you give some more input?

PR #130, opened 2023-10-17
add C and python API calls for solvation

### LINKED ISSUE #129: Solvent and GFN-FF support
Hi All,

After realizing TBLite is now the recommended upgrade from xtb-python, I noticed two features that can play a prominent role in my existing workflows are missing from TBLite that were implemented in xtb-python (particularly in the ASE interface) that I haven't seen another issue raised on: 1. implicit solvent effects 2. GFN-FF support. It would seem that adding support for 1 should be easier, but 2 might require a different library interface. Are there any plans to implement these features?

Cheers,
Michael.

pitsteinbach:
Adding implicit solvation models ALPB and CPCM to the C and python API, as requested by #129.

### DIFF
--- a/include/tblite/container.h
+++ b/include/tblite/container.h
@@ -50,6 +50,27 @@ tblite_new_spin_polarization(tblite_context ctx,
                              tblite_calculator calc,
                              double wscale);
 
+#define tblite_new_cpcm_solvation(ctx, mol, calc, x)            \
+                        _Generic((x),                           \
+                                char*                           \
+                                : tblite_new_cpcm_solvation_str,\
+                                double                          \
+                                : tblite_new_cpcm_solvation_dbl \
+                                ) (ctx, mol, calc, x)
+
+/// Create new CPCM implicit solvation container using internal parameters
+///
+/// @param ctx: Context handle
+/// @param mol: Molecular structure data
+/// @param calc: Calculator instance
+/// @param eps: epsilon value for solvent
+/// @return New interaction container
+TBLITE_API_ENTRY tblite_container TBLITE_API_CALL
+tblite_new_cpcm_solvation_dbl(tblite_context ctx,
+                          tblite_structure mol,
+                          tblite_calculator calc,
+                          double eps);
+
 /// Create new CPCM implicit solvation container using internal parameters
 ///
 /// @param ctx: Context handle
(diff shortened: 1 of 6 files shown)


pitsteinbach:
Seems like there is a difference in the energies depending on the use of intel or GNU compilers with the CPCM model. It seems to happen only when casting string to double.

pitsteinbach:
> Seems like there is a difference in the energies depending on the use of intel or GNU compilers with the CPCM model. It seems to happen only when casting string to double.

I compiled a fresh version and got a different value for CrCp2 with CPCM and an epsilon value of 7.0, which is now consistent with the results I got from the C-API test. Would be happy to have somebody else checking these values.

src/tblite/api/container.f90
```diff
@@ -149,6 +152,125 @@ function new_spin_polarization_api(vctx, vmol, vcalc, wscale) result(vcont) &
...
+
+   solvent = get_solvent_data(solvinp)
+   if (solvent%eps <= 0.0_wp) then
+      read(solvinp, *, iostat=stat) solvent%eps
+      if (stat /= 0) then
+         call fatal_error(error, "String value for epsilon could not be read as a real number")
+         call ctx%ptr%set_error(error)
+         return
+      end if
+   end if
```
awvwgk: I don't think this is the right thing to do here, can we define separate APIs for real and string inputs?

pitsteinbach: I changed to separate routines for string and float input, I added a macro to the C header file and adapted the tests accordingly.

include/tblite/container.h
```diff
@@ -50,6 +50,73 @@ tblite_new_spin_polarization(tblite_context ctx,
...
+
+/// Create new CPCM implicit solvation container using internal parameters
+///
+/// @param ctx: Context handle
+/// @param mol: Molecular structure data
+/// @param calc: Calculator instance
+/// @param solvent: Solvent to be modelled, can be given as name of solvent or epsilon value
+/// @return New interaction container
+TBLITE_API_ENTRY tblite_container TBLITE_API_CALL
+tblite_new_cpcm_solvation_str(tblite_context ctx,
```
awvwgk: How about solvent instead of str?

include/tblite/container.h
```diff
@@ -50,6 +50,73 @@ tblite_new_spin_polarization(tblite_context ctx,
...
+
+/// Create new CPCM implicit solvation container using internal parameters
+///
+/// @param ctx: Context handle
+/// @param mol: Molecular structure data
+/// @param calc: Calculator instance
+/// @param eps: epsilon value for solvent
+/// @return New interaction container
+TBLITE_API_ENTRY tblite_container TBLITE_API_CALL
+tblite_new_cpcm_solvation_dbl(tblite_context ctx,
```
awvwgk: How about epsilon or dielectric instead of dbl?

awvwgk (approved):
Looks good to me

haneug:
Looks good! I only have two small things to add:
1. We could add the newly available interactions to the docstring of the add function in the Calculator class:
https://github.com/tblite/tblite/blob/0c289785ae505c58e264bd1438fad0d4f402dcc1/python/tblite/interface.py#L510-L519
2. and I found two small typos (cpcm instead of alpb):
https://github.com/tblite/tblite/blob/0c289785ae505c58e264bd1438fad0d4f402dcc1/src/tblite/api/container.f90#L250
and
https://github.com/tblite/tblite/blob/0c289785ae505c58e264bd1438fad0d4f402dcc1/src/tblite/api/container.f90#L292

I compared the results of the Python API with the command line for Intel compilers with CrCp2 and the results nicely matched!

The GBSA solvation model could also be added in the future but first the full ALPB and GBSA should be implemented I think (see #138)

pitsteinbach:
I changed the suggested lines, thanks for catching the typos.
Once you give approval, I will merge and close the PR.

haneug (approved):

PR #133, opened 2023-10-27
Refactor CEH model 

thfroitzheim:
The CEH model implemented in #124 was refactored to follow the layout for the GFN Hamiltonians more closely. This includes: 
1. Transfer of ceh_guess into its own singlepoint.f90 file.
2. Transfer of the Hamiltonian setup and the (rewritten) get_hamiltonian function into the h0.f90 file. In this course, the new ceh_hamiltonian and ceh_calculator classes were removed and replaced by the tb_hamiltonian and xtb_calculator base classes (added data will also be required in the GP3-xTB development). 
3. Introduction of a ceh_h0spec to use the tb_hamiltonian machinery for the Hamiltonian setup. Expansion of the getter functions needed for get_hamiltonian. 
4. Consistent renaming of all variables to match the prior GFN implementation. 
5. Expansion of the CEH-specific test suite. 
The restructuring is necessary to use the already available GFN implementation for the setup of analytical charge gradients of CEH. 

The coordination number was restructured to prevent code duplication due to the many different CNs (now renamed). A new CN now only requires defining the count function and its derivative. To allow directed CN functions (i.e. erf_en, added to one partner and subtracted from the other) a directed_factor is included (I am open to better ideas). In this course, the testing was expanded to cover all possible CN implementations. 

Lastly, the derivatives of the diatomic frame scaled overlap was implemented.


src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,251 @@
...
+
+!> @file tblite/ceh/singlepoint.f90
+!> Provides main entry point for performing a CEH calculation
+!> follows in close analogy the xtb singlepoint
+
+!> Implementation of the single point calculation for the CEH model
+module tblite_ceh_singlepoint
+   !> mctc-lib
+   use mctc_env, only : error_type, wp
+   use mctc_io, only: structure_type
```
awvwgk: The comments in the import part do not add much explanation, please consider removing them, e.g. the comment `! Wavefunction` should be unnecessary for a module called `use tblite_wavefunction`. A more useful way of commenting is to describe why or how the statement or declaration is used, e.g. `use tblite_basis_type, only : get_cutoff  ! Realspace cutoff for integral calculation`.

src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,251 @@
...
+      type(adjacency_list) :: list
+      !> Potential type
+      type(potential_type) :: pot
+      !> Restart data for interaction containers
+      type(container_cache) :: icache
+      !> Timer
+      type(timer_type) :: timer
+      real(wp) :: ttime
+
+      logical :: grad = .false.
```
awvwgk: This variable is implicitly saved, this is a bug.

src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,251 @@
...
+
+      real(wp) :: elec_entropy
+      real(wp) :: nel = 0.0_wp, cutoff
```
awvwgk: Also, this variable is implicitly saved, this is generally not the wanted behavior.

src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,251 @@
...
+            & "   " // format_string(wfn%qat(i,1), real_format))
+         end do
+         call ctx%message(repeat("-", 60))
+         call ctx%message(label_dipole)
+         call ctx%message("     x           y           z")
+         call ctx%message(format_string(dipole(1), "(f12.5)") // &
+         & format_string(dipole(2), "(f12.5)") // &
+         & format_string(dipole(3), "(f12.5)"))
+         call ctx%message(repeat("-", 60))
+      endif
```
awvwgk: Is this printout necessary here or can it be implemented on a higher level, e.g. in the guess or main driver?

src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,251 @@
...
+   end subroutine ceh_guess
+
+   subroutine header(ctx)
+      !> Calculation context
+      type(context_type), intent(inout) :: ctx
+      call ctx%message(repeat("-", 60))
+      call ctx%message('       C H A R G E    E X T E N D E D    H U C K E L (CEH) ')
+      call ctx%message('                       SG, MM, AH, TF, May 2023            ')
+      call ctx%message(repeat("-", 60))
+   end subroutine header
```
awvwgk: tblite is by design free of headers.

src/tblite/ncoord/erf.f90
```diff
@@ -0,0 +1,137 @@
...
+      exponent = kcn*(r-rc)/rc
+
+      expterm = exp(-exponent**2)
+
+      count = -(kcn*expterm)/(rc*sqrt(pi))
+
+
+   end function ncoord_dcount
+
+
```
awvwgk: Please check whether the extra empty lines are necessary here.

src/tblite/ncoord/exp.f90
```diff
@@ -18,73 +18,121 @@
 !> Provides a coordination number implementation with exponential counting function
 
 !> Coordination number implementation using an exponential counting function
+!> Same as in dftd3 reimplemented
 module tblite_ncoord_exp
    use mctc_env, only : wp
    use mctc_io, only : structure_type
-   use tblite_cutoff, only : get_lattice_points
+   !use tblite_cutoff, only : get_lattice_points
```
awvwgk: Please removed commented out code.

awvwgk (changes requested):
Thanks for taking the effort of merging the CEH hamiltonian with the xTB one, this is a quite positive change. I left a few comments on issues and bugs which I could spot while skimming through the PR.

src/tblite/ceh/singlepoint.f90
thfroitzheim: That is good to know. I removed all cases, where it happened

src/tblite/ceh/singlepoint.f90
thfroitzheim: No that is not needed. I remove it.

thfroitzheim:
I fixed all of the formatting issues you pointed out and removed unnecessary output and the implicitly saved variable definitions. In this course, I updated also the naming of the tests as now many versions of ncoord are tested, which led to a small merge conflict (I think I resolved it correctly).

It took a bit longer because there was already the first update for the CEH parameters. I updated all parameters and moved the radii and EN values into the ceh.f90 file, as they are also fitted. Otherwise, there was no major change to the code.

app/cli.f90
```diff
@@ -429,6 +433,12 @@ subroutine get_run_arguments(config, list, start, error)
          call get_argument_as_real(arg, config%etemp, error)
          if (allocated(error)) exit
 
+      case("--etemp_guess")
```
awvwgk: Let's use dashes rather than underscores.

src/tblite/ceh/h0.f90
```diff
@@ -16,18 +16,292 @@
...
+      if (present(dsedcn)) dsedcn(:) = 0.0_wp
+      if (present(dsedcn_en)) dsedcn_en(:) = 0.0_wp
+      do iat = 1, size(id)
+         izp = id(iat)
+         ii = ish_at(iat)
+         do ish = 1, nshell(izp)
+            selfenergy(ii+ish) = h0%selfenergy(ish, izp)
+         end do
+      end do
+      if (present(cn) .and. present(cn_en)) then
```
awvwgk: Can this be
```suggestion
      if (.not.present(cn) .or. .not.present(cn_en)) return
```
instead to reduce the level of indentation. Should also make the code structure somewhat clearer.

src/tblite/ceh/singlepoint.f90
```diff
@@ -0,0 +1,207 @@
...
+
+      logical :: grad
+
+      real(wp) :: elec_entropy
+      real(wp) :: nel, cutoff
+      real(wp), allocatable :: tmp(:)
+
+      integer :: i, prlevel
+
+      !> coordination number related arrays
```
awvwgk: Please don't use docstring markers for non-docstring comments. These comments are prone to end up in the wrong place when extracted by ford or doxygen. Only dummy variables or procedures should have comments with the docstring marker `>`.

src/tblite/ncoord/exp.f90
```diff
@@ -18,73 +18,106 @@
 !> Provides a coordination number implementation with exponential counting function
 
 !> Coordination number implementation using an exponential counting function
+!> Same as in dftd3 reimplemented
```
awvwgk: Does this need a dot or comma at the end of the previous sentence?

awvwgk (approved):
Only some minor comments, now it looks like in a good shape to get merged.

src/tblite/ceh/singlepoint.f90
thfroitzheim: I have found a few more and removed them.

src/tblite/ceh/h0.f90
thfroitzheim: That is a better way.

thfroitzheim:
Great. I have fixed the few points you mentioned and also updated the help page for the guess.

awvwgk:
Can you please also make sure the entries in man/ are consistent with the help output?

thfroitzheim:
I have added a new page to man for the guess and the guess temperature to the page for run.

awvwgk (approved):
Looks good to me now

thfroitzheim:
I think that you will have to merge the PR as I am not a maintainer (there is no hurry, but you don't have to wait for me).

PR #135, opened 2023-11-05
Add error handling to callback API for logger

### LINKED ISSUE #134: The tblite ASE calculator interface does not properly handle `KeyboardException` (Ctrl+C)
When running a structure relaxation using ASE and tblite, doing Ctrl+C (keyboard interrupt) cancels the current cycle but not the run as a whole. This makes it very difficult and inconvenient to abort a run.

```python
from tblite.ase import TBLite
from ase.build import bulk
from ase.optimize import BFGS

atoms = bulk("Cu")*(3,3,3)
atoms.calc = TBLite()
dyn = BFGS(atoms)
dyn.run(fmax=1e-4)
```

Then Ctrl+C before it finished. Tested on tblite 0.3.0.

image

awvwgk:
Closes #134

### DIFF
--- a/python/tblite/library.py
+++ b/python/tblite/library.py
@@ -86,7 +86,7 @@ def logger_callback(error, message, nchar, data):
         callback = ffi.from_handle(data)
         callback(ffi.unpack(message, nchar).decode())
     except Exception as e:
-        error_message = ffi.new("char[]", str(e))
+        error_message = ffi.new("char[]", str(e).encode())
         lib.tblite_set_error(
             error,
             error_message,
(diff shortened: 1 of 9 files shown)


python/tblite/library.py
```diff
@@ -80,11 +80,18 @@ def handle_error(*args, **kwargs):
...
+def logger_callback(error, message, nchar, data):
     """Custom logger callback to write output in a Python friendly way"""
-
-    callback = ffi.from_handle(data)
-    callback(ffi.unpack(message, nchar).decode())
+    try:
+        callback = ffi.from_handle(data)
+        callback(ffi.unpack(message, nchar).decode())
+    except Exception as e:
+        error_message = ffi.new("char[]", str(e))
```
awvwgk:
```suggestion
        error_message = ffi.new("char[]", str(e).encode())
```

PR #142, opened 2023-12-05
Full ALPB/GBSA with empirical parameters

### LINKED ISSUE #138: Different solvation energies in tblite and xtb
Hi together,

I have noticed that application of both ALPB and GBSA solvation models resulted in different energies for the newest versions of tblite (the installed executable after building with meson) and xtb.
A simple example is a single point calculation of ethanol (ethanol.xyz):

    9    
    Ethanol
      H      1.8853     -0.0401      1.0854
      C      1.2699     -0.0477      0.1772
      H      1.5840      0.8007     -0.4449
      H      1.5089     -0.9636     -0.3791
      C     -0.2033      0.0282      0.5345
      H     -0.4993     -0.8287      1.1714
      H     -0.4235      0.9513      1.1064
      O     -0.9394      0.0157     -0.6674
      H     -1.8540      0.0626     -0.4252

Without solvation, both programs essentially give the same energy (as expected):

    tblite ethanol.xyz

resulted in:

    ....
    ------------------------------------------------------------
      cycle        total energy    energy error   density error
    ------------------------------------------------------------
          1     -11.34484894099  -1.1462750E+01   1.6126437E-01
          2     -11.38501140154  -4.0162461E-02   7.2796973E-02
          3     -11.39038158210  -5.3701806E-03   2.0933834E-02
          4     -11.39068145220  -2.9987011E-04   1.0279824E-02
          5     -11.39074527592  -6.3823713E-05   1.8471074E-03
          6     -11.39075016731  -4.8913914E-06   9.5529263E-04
          7     -11.39075043232  -2.6501404E-07   3.7627198E-04
          8     -11.39075091328  -4.8096264E-07   1.4263204E-04
          9     -11.39075096916  -5.5874589E-08   9.6098618E-06
    ------------------------------------------------------------

    electronic energy        -1.1508651630483E+01 Eh
    total energy             -1.1390750969159E+01 Eh
    ....

and 

    xtb ethanol.xyz

resulted in:

               -------------------------------------------------
              | TOTAL ENERGY              -11.390750963714 Eh   |
              | GRADIENT NORM               0.185597446632 Eh/α |
              | HOMO-LUMO GAP              12.836665182229 eV   |
               -------------------------------------------------

When applying ALPB with water as solvent, however, the results are different:

    tblite ethanol.xyz --alpb water

gives 

    -11.39322361699 Eh

but 

    xtb  ethanol.xyz --alpb water

gives

    -11.397759549872 Eh

Thus, the solvation energy seems more than twice as large for the xtb program as in the case of tblite, even if the same structure is calculated with the same solvation model. Similar deviations can be observed by using GBSA.

I therefore wanted to ask if there are some different default settings for both programs which must be changed by the user to get the same results or if there is some issue with the solvation in tblite in general.

Thanks in advance!

haneug:
Initial implementation of the 'full' ALPB and GBSA solvation models as in xtb. 

- implements new flags --alpb-xtb and --gbsa-xtb which employ the corresponding empirical parameters and give the same results as in xtb (for water in water the energy difference is below 10⁻⁸ Eh with GFN2-xTB)
- No support for API yet
 
Looking forward for your suggestions how to improve this! 

To Do:
- [x] implement Gshift
- [x] fix deviation for GFN1-xTB 
- [x] check other solvents
- [x] clean up and add tests
- [x] C and python api


app/driver_run.f90
```diff
@@ -61,7 +61,7 @@ module tblite_driver_run
 
 
 subroutine run_main(config, error)
-   type(run_config), intent(in) :: config
+   type(run_config), intent(inout) :: config
```
awvwgk: This should stay intent in

haneug: Definitely, I want to set the parameters in new_solvation because there I have mol available. My quick (and dirty?) solution is to use a scratch_input in new_solvation so that I do not have to modify config%solvation%alpb.

haneug:
Differences for GFN1-xTB do not occur for molecules with only one species e.g., H2, F2, and O3, but for molecules with multiple species e.g., HF, H2O, and CO2.

pprcht:
Hi all,

To put this issue a bit forward, I committed some changes to @haneug's branch. I think this is an important PR, so I'm happy to further work on it.

Here's what I did:
- Updated the branch to the current main branch stage.
- Updated solvent strings. Some solvents like `dmf`, `thf`, `cs2` and a few others were not available under these names, but could be selected under their full name. `tblite` now knows the alias as well. `acetone` was missing entirely, I added the entry with parameters from MNSOL.
- The solvents `furane`, `benzaldehyde`, `dioxane`, `phenol` and `woctanol` were not available in `tblite` *nor* in MNSOL. I added their entries to the `get_solvent_data` constructor, but they only contain the dielectric constant and refractory indices.
- The discrepancy between GFN1-xTB energies stemmed from a CM5 charge implementation in `xtb` that was missing in `tblite`. Not anymore.
- Finally, I tested all GFN1/GFN2 GBSA/ALPB combinations with all available solvents. The test molecule was caffeine:
```
$coord
         -6.134194835085         -2.050583457383          0.041501488892     c
         -4.257989774446         -0.077573907965         -0.001827990178     n
         -4.712442263975          2.470621717876         -0.013111881308     c
         -2.592957527478          3.781873878383         -0.016386752339     n
         -0.735810404277          1.998874697504         -0.012516600988     c
         -1.696825680417         -0.388511985183         -0.013341342571     c
         -0.165536422836         -2.593558889366         -0.012308491761     c
         -1.052950170836         -4.736438814102         -0.016475781913     o
          2.417331343252         -2.043834570361         -0.004348903618     n
          4.182092079482         -4.138750561195          0.008357048520     c
          3.486490066997          0.397636251409          0.000102411769     c
          5.802113920722          0.690657762313          0.005286648234     o
          1.822761570451          2.443307017589         -0.000190631983     n
          2.776162986238          5.014434390170          0.020533754716     c
         -8.020562503708         -1.210549395752         -0.004021712955     h
         -5.897943375362         -3.139351862292          1.780647162725     h
         -5.867941387821         -3.252537075648         -1.616202723931     h
         -6.598840021510          3.263573033853         -0.013105328556     h
          3.236734851046         -5.974575385829          0.004320331970     h
          5.372500093097         -4.001905151256          1.695285118496     h
          5.396766837421         -4.005607709880         -1.661375665537     h
          4.840892905022          5.070245799787          0.070744083415     h
          2.052500009863          5.991405353355          1.694401414324     h
          2.131182261180          5.989581702181         -1.685866648900     h
$end
```
Here are the summarized results:
[gfn2-gbsa-results.txt](https://github.com/user-attachments/files/16641356/gfn2-gbsa-results.txt)
[gfn2-alpb-results.txt](https://github.com/user-attachments/files/16641357/gfn2-alpb-results.txt)
[gfn1-gbsa-results.txt](https://github.com/user-attachments/files/16641358/gfn1-gbsa-results.txt)
[gfn1-alpb-results.txt](https://github.com/user-attachments/files/16641360/gfn1-alpb-results.txt)

As Hagen mentioned in the beginning, the deviation in total energy w.r.t. the `xtb` implementation was on the order of ~10<sup>-8</sup> Eh, and it still is, now for all solvents and both Hamiltonians. The gradients match pretty well, too.

### TODO
As I said above, I'm happy to continue working on this and there may be still some things to do. It would be great if you advise me on how/what you would like for the implementation.
- add some tests
- More cleanup? E.g. should setting this implementation be kept as `--alpb-xtb`/`--gbsa-xtb` or would you prefer it to be just `--alpb`/`--gbsa`? The latter may be preferable to avoid confusion...

pprcht:
Added some tests and fixed some of the logic of CM5 charge allocation

pitsteinbach:
I added the python and C-API, for the changes done by the other guys and added some tests to go along.
I had to add a string to the calculator containing the method to retrieve this information in the container API, sop that the right parameters can be chosen.

thfroitzheim:
This should now be ready for review. It is tested for all solvents against xtb. Most remaining misses in code coverage are in the CLI for which there are additional test.

include/tblite/container.h
```diff
@@ -84,12 +85,20 @@ tblite_new_cpcm_solvation_solvent(tblite_context ctx,
                           tblite_calculator calc,
                           char* solvent);
 
-#define tblite_new_alpb_solvation(ctx, mol, calc, x)            \
+// standard/default argument for reference state of solvation model
+char * std_refstr = "gsolv";
+//Macro to call tblite_new_alpb_solvation with or without/default reference string
+#define tblite_new_alpb_solvation(...) vrg(tblite_new_alpb_solvation, __VA_ARGS__)
```
awvwgk: Why is it necessary to have variadic functions here. Let's define two separate and clean APIs and combine them with a type generic macro instead.

Keep in mind eventually we need to expose the ABI to Python and any macro trick we are using in C is not possible to export there.

include/tblite/container.h
```diff
@@ -84,12 +85,20 @@ tblite_new_cpcm_solvation_solvent(tblite_context ctx,
                           tblite_calculator calc,
                           char* solvent);
 
-#define tblite_new_alpb_solvation(ctx, mol, calc, x)            \
+// standard/default argument for reference state of solvation model
+char * std_refstr = "gsolv";
```
awvwgk: Instead of using a global variable here we can define an enumerator for the reference state instead. If passed as int* a NULL would show absence of the optional argument. In Python we can add a wrapper to convert strings to enums to make it more user-friendly.

awvwgk (changes requested):

include/tblite/container.h
pitsteinbach: Well I wanted to keep the API for ALPB and GBSA constant and identical to that of CPCM, default values are not possible in C and from what I could find on the internet variadic functions are the way to go.
I don't see how to combine using a Generic Macro since from what I can find, Generic must have an operand which would not be possible if you call new_alpb_solvation with only 4 arguments.

I don't understand your second point, since Generic is also a macro trick or isn't it? Maybe you can point me to some further reading.

include/tblite/container.h
pitsteinbach: I proposed a solution using an enum and `int*` API, which impacted the general ALPB/GBSA API. I also propose a python wrapper for the enum I don't know if that is what you expected.

app/tester.py
awvwgk, line 1: Can we move this to `test/validation` to keep the app directory clean with only the files relevant for distribution?

app/cli.f90
```diff
@@ -357,24 +361,65 @@ subroutine get_run_arguments(config, list, start, error)
...
             call get_argument_as_real(arg, solvent%eps, error)
          end if
          if (allocated(error)) exit
+
+         ! Check for optional solution state 
+         sol_state = solutionState%gsolv
+         iarg = iarg + 1
+         call list%get(iarg, arg)
+         if (allocated(arg)) then
+            if (arg == "gsolv") then
```
awvwgk: Let's add a new option called `--solvation-state` or similar instead of making the options for the solvation model more involved.

awvwgk (approved):

awvwgk:
Thanks Hagen @haneug, Philipp @pprcht, Thomas @thfroitzheim and Pit @pitsteinbach for the all the hard work on this.

PR #145, opened 2023-12-20
Update of the spGFN spin constants

benbaed:
Updated the rare gas spin constants as they were previoulsy mistakenly computed using HF.
Also updated Barium (added d constants)


awvwgk (changes requested):
Could you please check whether the patch only changes the rare gases and barium, since the diff is quite large.

PR #154, opened 2024-01-19
Fix OMP memory leak in Coulomb matrix derivatives

pprcht:
Many thanks for this awesome software!

I've noticed an apparent leak of memory that occurs for a large number of energy+gradient evaluations (e.g. during long MD simulations or extensive sampling applications). 
For a handful of e+grd calls one barely notices it, but the memory consumption becomes severe over time, leading to swapping and eventually blows up the call:
memory

## How to reproduce?
The error can be easily reproduced with the `tblite` app, e.g. by simply building a loop around the `xtb_singlepoint` call. To reproduce it the gradient calculation (`--grad`) must be requested, a simple SCC without gradient does *not* show this.
Concerning technical setting: 
I've been using the latest  commit from the main branch (#152) and the build was created using
**meson 1.2.0** and **ifort 2021.9.0** with an MKL backend from the same oneAPI verison. 
The build was a static one, i.e., using `--default-library=static`, as well as Fortran link arguments `-static` and `-qopenmp`, but otherwise used the default build options.

## What was changed?
I've traced the memory consumption back to the OMP-parallel evaluation of the Coulomb matrix derivatives. My interpretation of what happens is that the implicit allocation of
```
   itrace = atrace
   didr = dadr
   didL = dadL
```
***within*** the OMP statement does not free up the respective memory correctly after execution of the subroutines `get_damat_3d` and `get_damat_0d`. This could be a problem endemic to the `ifort` build, but I'm not 100% sure about that. 
Allocating the three tensors *before* the OMP statement resolves the issue and even a few hundred thousand energy and gradient calls do not increase the memory consumption. 
memory2

All builds and unit tests are passing, there is no other internal or external change to the behavior of the code. :+1:

### DIFF
--- a/src/tblite/coulomb/charge/effective.f90
+++ b/src/tblite/coulomb/charge/effective.f90
@@ -541,13 +541,13 @@ subroutine get_damat_0d(mol, nshell, offset, hubbard, gexp, qvec, dadr, dadL, at
    dadr(:, :, :) = 0.0_wp
    dadL(:, :, :) = 0.0_wp
 
+   itrace = atrace
+   didr = dadr
+   didL = dadL
    !$omp parallel default(none) &
    !$omp shared(atrace, dadr, dadL, mol, qvec, hubbard, nshell, offset, gexp) &
    !$omp private(iat, izp, ii, ish, jat, jzp, jj, jsh, gam, r1, vec, dG, dS, dtmp, arg) &
    !$omp private(itrace, didr, didL)
-   itrace = atrace
-   didr = dadr
-   didL = dadL
    !$omp do schedule(runtime)
    do iat = 1, mol%nat
       izp = mol%id(iat)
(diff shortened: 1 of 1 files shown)


awvwgk (changes requested):
Can we move the sourced allocation in the omp region? This allocation should be done by each thread and not in a sequential region and then broadcast. Maybe we can explicitly deallocate the arrays after the omp loop.

pprcht:
> Can we move the sourced allocation in the omp region? This allocation should be done by each thread and not in a sequential region and then broadcast. Maybe we can explicitly deallocate the arrays after the omp loop.

Yes, makes sense. Let me check!

pprcht:
Explicit allocation/deallocation within the OMP region does it.

src/tblite/coulomb/charge/effective.f90
```diff
@@ -579,6 +582,7 @@ subroutine get_damat_0d(mol, nshell, offset, hubbard, gexp, qvec, dadr, dadL, at
    dadr(:, :, :) = dadr + didr
    dadL(:, :, :) = dadL + didL
    !$omp end critical (get_damat_0d_)
+   deallocate(didL,didr,itrace)
```
awvwgk:
```suggestion
   deallocate(didL, didr, itrace)
```

src/tblite/coulomb/charge/effective.f90
```diff
@@ -684,6 +691,7 @@ subroutine get_damat_3d(mol, nshell, offset, hubbard, gexp, rcut, wsc, alpha, qv
    dadr(:, :, :) = dadr + didr
    dadL(:, :, :) = dadL + didL
    !$omp end critical (get_damat_3d_)
+   deallocate(didL,didr,itrace)
```
awvwgk:
```suggestion
   deallocate(didL, didr, itrace) 
```

awvwgk (approved):

awvwgk (changes requested):
Sourced allocation should be identical to automatic LHS allocation on assignment, can we remove the either the assignment or the allocation? The main issue of the leak seems to be the missing deallocation.

pprcht:
> Sourced allocation should be identical to automatic LHS allocation on assignment, can we remove the either the assignment or the allocation? The main issue of the leak seems to be the missing deallocation.

It is. I left the sourced allocation and removed the LHS assignment. This way it is slightly more obvious where the corresponding deallocation comes from.

awvwgk (approved):

PR #157, opened 2024-02-05
Bug fix double dictionary testsuite & removal of entries.

pitsteinbach:
The initial implementation of the double dictionary had some bugs regarding the removal of entries. The last entry instead of the one matching the label was removed.
This is fixed now, I adapted the testsuite to now capture such a bug.
I further extended the functionality to a toml read in as well as write functionality which is covered by tests.
Inspired by those tests I added an equal operator which is tested as well.

Now the dictionary should work as intended, sorry for the faulty first version.


test/unit/test_double_dictionary.f90
```diff
@@ -524,4 +555,122 @@ subroutine test_update_entries_label(error)
...
+
+subroutine test_equal_operator_different_dict(error)
+   type(error_type), allocatable, intent(out) :: error
+   type(double_dictionary_type) :: dict, dict1, dict2, dict3
+   real(wp), allocatable :: array1(:)
+   logical :: che
+   allocate(array1(4), source = 1.0_wp)
+   call fill_test_dict(dict)
+   call fill_test_dict_1d_array(dict1)
+   che = (not_(dict == dict1))
```
awvwgk: Why can't you use `.not.` here?

awvwgk:
Could you separate the bug fix from the new features? This would make it much easier to review.

test/unit/test_double_dictionary.f90
pitsteinbach: You are totally right, I somehow forgot about it but was using not() instrinsic instead.

pitsteinbach:
Yes sorry I thought that since it is only small changes I would add them. 
This PR should handle the Bug Fix then.

awvwgk (approved):

PR #158, opened 2024-02-08
Add toml read and write functionality to double dictionary.

pitsteinbach:
Added toml read and write capability to double dictionary class.

This can be used to dump post processing results for example to toml files which are use commonly now in python applications. The toml functionality previously used in the project is thereby further expanded to the double dictionary type.

Directly using a toml table pointer to retrieve entries in python for example could be imagined.


awvwgk (commented):
With this mapping from and to TOML, what is the main advantage of reimplementing the data structures with a different API compared to TOML-Fortran? I can see only the stronger typing as an advantage here, but TOML-Fortran does provide an API to fetch into an array already.

pitsteinbach:
I am not sure what is already implemented by TOML-Fortran.

I added this since I wanted to dump the double dictionary let's say from an xtbml calculation to a toml file, if xtbml is requested from CLI. I tought that using toml has the advantage of already having the functionality of dumping arrays and is very compatible with python for further use in ML scripts or so.

awvwgk:
A `toml_table` does implement most of the functionality of the double dictionary already, without the type constraint.

pitsteinbach:
I see your point.
The double dictionary can store arrays with more dimensions easily and its OOP structure is in my opinion more convenient. Also combining instances of dictionaries easily is a plus for me.

We should just decide how to move on from here, we can dissect the dictionary out again, which involves reworking the post processing PR mainly, and then further implementations I have planned.

awvwgk (approved):

PR #162, opened 2024-02-11
Add support for pyberny optimizer

awvwgk:
Implement generator for consuming geometries from berny optimizer and sending energy and gradients.

ping Jan @jhrmnn


awvwgk:
Closing this for now. Might revisit this in the future as pyberny is now advertised in the Python docs.

PR #170, opened 2024-02-27
Python-interface: Loading spin-polarized quantities

pro7onmatrix:
Currently (latest commit is https://github.com/tblite/tblite/commit/5b0410da76cb2dc9a1fad41d332f0a12eeb091b5), the Python interface segfaults when trying to load quantities such as the density matrix in spin-polarized calculations.

For example, the following would trigger the segmentation fault:
```python
import numpy as np
from tblite.interface import Calculator


def main():
    numbers = np.array([1, 8])
    positions = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0]])

    calc = Calculator("GFN1-xTB", numbers, positions)
    calc.add("spin-polarization")
    res = calc.singlepoint()
    p = res.get("density-matrix")


if __name__ == "__main__":
    main()

```

This PR fixes that bug and extends the Python interface to handle spin-polarized quantities in the following way:
- The function `tblite_get_result_number_of_orbitals_api` now returns the number of spatial orbitals instead of the number of spin orbitals
- A new function `tblite_get_result_number_of_spins_api` returns the number of spins (1 or 2)
- If `spin-polarized` is set and a spin-polarized quantity is loaded, the resulting array will have the shape `(2, norb)` or `(2, norb, norb)`. This allows convenient destructuring, e.g.,

    ```python
    occs_a, occs_b = res.get("orbital-occupations")  # occs_a.shape == occs_b.shape == (norb,)
    den_a, den_b = res.get("density-matrix")  # den_a.shape == den_b.shape == (norb, norb)
    ```

    If `spin-polarized` is not set or the quantity is spin-independent (e.g., the overlap matrix), the spin-axis will be removed and the array has dimensions `(norb,)` or `(norb, norb)` as usual.

I have run the complete test suite and also added a Python and C-API test of my own for the changes made. Please let me know if you need any additional info or if there are any problems with the changes.


pro7onmatrix:
I overlooked that MO coefficients are returned transposed (i.e., with their MO index as the first axis).

While one might solve this by explicitly defining the matrices to be in column-major order as
```python
_mat = np.zeros((_norb, _norb, _nden), order="F")
```
this would make the assignment of alpha and beta matrices a bit awkward to write:
```python
orbitals = res.get("orbital-coefficients")
orbitals_a = orbitals[..., 0]
orbitals_b = orbitals[..., 1]
```
I instead opted to just swap the axes within `_get_ao_matrix` using [`np.swapaxes`](https://numpy.org/doc/stable/reference/generated/numpy.swapaxes.html).

python/tblite/interface.py
```diff
@@ -262,26 +262,26 @@ def get(self, attribute: str):
...
-         orbital-occupations    norb        e
-         orbital-coefficients   norb        unitless
-         overlap-matrix         norb, norb  unitless
-         hamiltonian-matrix     norb, norb  Hartree
-         density-matrix         norb, norb  e
-         natoms                 scalar      unitless
-         norbitals              scalar      unitless
-        ====================== =========== ==============
+        ====================== ============================= ==============
+         property               dimension                     unit
```
pitsteinbach:
```suggestion
         property               dimension  [spin-polarized case]        unit
```

python/tblite/interface.py
```diff
@@ -262,26 +262,26 @@ def get(self, attribute: str):
...
+        ====================== ============================= ==============
+         energy                 scalar                        Hartree
+         energies               nat                           Hartree
+         gradient               nat, 3                        Hartree/Bohr
+         virial                 3, 3                          Hartree
+         charges                nat                           e
+         bond-orders            nat, nat                      e
+         dipole                 3                             e·Bohr
+         quadrupole             6                             e·Bohr²
+         orbital-energies       norb or 2, norb               Hartree
```
pitsteinbach:
```suggestion
         orbital-energies       norb [2, norb]               Hartree
```

python/tblite/interface.py
```diff
@@ -262,26 +262,26 @@ def get(self, attribute: str):
...
+         orbital-occupations    norb or 2, norb               e
```
pitsteinbach:
```suggestion
         orbital-occupations    norb [2, norb]               e
```

python/tblite/interface.py
```diff
@@ -262,26 +262,26 @@ def get(self, attribute: str):
...
+         orbital-coefficients   norb, norb or 2, norb, norb   unitless
```
pitsteinbach:
```suggestion
         orbital-coefficients   norb, norb  [2, norb, norb]   unitless
```

python/tblite/interface.py
```diff
@@ -262,26 +262,26 @@ def get(self, attribute: str):
...
+         overlap-matrix         norb, norb                    unitless
+         hamiltonian-matrix     norb, norb                    Hartree
+         density-matrix         norb, norb or 2, norb, norb   e
```
pitsteinbach:
```suggestion
         density-matrix         norb, norb [2, norb, norb]   e
```

test/api/main.c
pitsteinbach, line 1: Thanks for taking care of the additional whitespace that I left there;)

pitsteinbach (commented):
Looks good, maybe add another test to capture empty wavefunction. I added a suggestion for the docstring to have a more clear layout to distinguish spin-polarized cases.

pro7onmatrix:
I'm not 100% sure what you mean by "empty wavefunction", could you give an example? Do you mean fully spin-polarized systems (e.g., hydrogen atom) where there is no (meaningful) beta-density?

test/api/main.c
pitsteinbach, line 1: Sorry I meant implement a test for an uninitialized result container, i.e. after line 394 after your additions.
Same for an initialized result container without actual results, so after line 673. Just to catch that one missing line in the API.

pro7onmatrix: I see, sorry for the confusion. I will add the tests. On that note, isn't the function `test_empty_result` kind of redundant, since `test_uninitialized_result` already performs all the same checks (and quite a few more) from line 300 onward?

pitsteinbach: I stumbled over that as well this morning, seems like I added these tests again form line 300 and onward, since I didn't see `test_empty_results`, which was there before according to git blame. I will open a PR to remove those redundant tests again.

PR #173, opened 2024-03-25
Memory Leak behavior of tblite when using MKL as LAPACK and BLAS backend

pitsteinbach:
We recently observed that there is a build up of memory usage when computing single points using `tblite` via the python API on varying molecules. When staying with one system size there is no build-up, however when changing molecules and thereby system size a lot, memory usage is going up.
We used a simple python script with a memcheck decorator to trace memory usage. I attached the initial output as `no_free.txt`. I realized that there is no build-up when using OpenBLAS as backend. Thereby suspecting that MKL has something to do with it.

I added a single call of `mkl_free_buffers` after the singlepoint routine, using a precompiler flag to compile the call when MKL is used. This change helped to keep the memory build up down. See `w_free.txt`.

[w_free.txt](https://github.com/tblite/tblite/files/14747066/w_free.txt)
[no_free.txt](https://github.com/tblite/tblite/files/14747067/no_free.txt)

### DIFF
--- a/config/meson.build
+++ b/config/meson.build
@@ -78,7 +78,7 @@ if lapack_vendor == 'mkl'
 elif lapack_vendor == 'mkl-rt'
   mkl_dep = fc.find_library('mkl_rt')
   lib_deps += mkl_dep
-  add_project_arguments(['-DW_MKL'], language: 'fortran')
+  add_project_arguments(['-DWITH_MKL'], language: 'fortran')
 elif lapack_vendor == 'openblas'
   openblas_dep = dependency('openblas', required: false)
   if not openblas_dep.found()
--- a/config/meson.build
+++ b/config/meson.build
@@ -74,11 +74,11 @@ if lapack_vendor == 'mkl'
   endif
   mkl_dep += cc.find_library('mkl_core')
   lib_deps += mkl_dep
-
+  add_project_arguments(['-DW_MKL'], language: 'fortran')
 elif lapack_vendor == 'mkl-rt'
   mkl_dep = fc.find_library('mkl_rt')
   lib_deps += mkl_dep
-
+  add_project_arguments(['-DW_MKL'], language: 'fortran')
 elif lapack_vendor == 'openblas'
   openblas_dep = dependency('openblas', required: false)
   if not openblas_dep.found()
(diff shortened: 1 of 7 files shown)


awvwgk (commented):
Can this be done as part of deconstructing the eigenvalue solver?

pitsteinbach:
yep works as well, I have moved the call to the solver deletion routine.

pitsteinbach:
I found an issue, if the post-processing containers or gradients use BLAS then it will reaccumulate buffer. But we could move the call to destroy the solver to the end of the singlepoint routine.

awvwgk:
I wonder what will happen if we deploy tblite via conda-forge where the BLAS implementation is determined at runtime, in this case we would be unable to free the accumulated buffers at all. Is this issue specific to a certain MKL version?

pitsteinbach:
This behavior can be observed for `oneMKL` 2023.2 and 2024.0, I tested for those versions explicitly, although I have access to more versions.

config/meson.build
```diff
@@ -74,11 +74,11 @@ if lapack_vendor == 'mkl'
   endif
   mkl_dep += cc.find_library('mkl_core')
   lib_deps += mkl_dep
-
+  add_project_arguments(['-DW_MKL'], language: 'fortran')
```
awvwgk:
```suggestion
  add_project_arguments(['-DWITH_MKL'], language: 'fortran')
```

config/meson.build
```diff
@@ -74,11 +74,11 @@ if lapack_vendor == 'mkl'
...
 elif lapack_vendor == 'mkl-rt'
   mkl_dep = fc.find_library('mkl_rt')
   lib_deps += mkl_dep
-
+  add_project_arguments(['-DW_MKL'], language: 'fortran')
```
awvwgk:
```suggestion
  add_project_arguments(['-DWITH_MKL'], language: 'fortran')
```

src/tblite/context/type.F90
```diff
@@ -153,7 +153,9 @@ subroutine delete_solver(self, solver)
    if (allocated(self%solver)) then
       call self%solver%delete(solver)
    end if
-
+#if defined W_MKL
```
awvwgk: Maybe that better goes in the beginning of the module

```suggestion
#ifndef WITH_MKL
#define WITH_MKL 0
#endif
#if WITH_MKL
```

awvwgk (approved):

pitsteinbach:
looks good

Issue #104, opened 2023-03-28
Inlined transformation matrix uses wrong orbital ordering


marvinfriede:
If one uses the [transformation matrix](https://github.com/tblite/tblite/blob/b3dcf324366eb053d41bd15af15b4a2d26487a4a/src/tblite/integral/trafo.f90#L33) for the transformation from cartesian to spherical basis functions, many tests throw an error. To be precise, using `sphr = matmul(dtrafo, matmul(cart, transpose(dtrafo)))` does not yield the same result as the [explicit transformation](https://github.com/tblite/tblite/blob/b3dcf324366eb053d41bd15af15b4a2d26487a4a/src/tblite/integral/trafo.f90#L135).
If I change the ordering from `[-l, ..., 0, ..., l]` to `[0, ..., l-1, -(l-1), l, -l]`, it works again.

```fortran
real(wp), parameter :: dtrafo(5, 6) = reshape([&
   !  0       1       -1      2       -2
   & -0.5_wp, 0.0_wp, 0.0_wp,   s3_4, 0.0_wp, & ! xx
   & -0.5_wp, 0.0_wp, 0.0_wp,  -s3_4, 0.0_wp, & ! yy
   &  1.0_wp, 0.0_wp, 0.0_wp, 0.0_wp, 0.0_wp, & ! zz
   &  0.0_wp, 0.0_wp, 0.0_wp, 0.0_wp,     s3, & ! xy
   &  0.0_wp,     s3, 0.0_wp, 0.0_wp, 0.0_wp, & ! xz
   &  0.0_wp, 0.0_wp,     s3, 0.0_wp, 0.0_wp],& ! yz
   & shape(dtrafo))
```

I suspect that the same applies to the transformation matrices for higher (than 2) angular momenta.

awvwgk:
Failing tests with explicit transformation matrices:

````
# Testing: integral-overlap
  Starting overlap-1 ... (1/11)
       ... overlap-1 [FAILED]
  Message: Floating point value missmatch
  Starting overlap-5 ... (5/11)
       ... overlap-5 [FAILED]
  Message: Floating point value missmatch
# Testing: hamiltonian
  Starting hamiltonian-3 ... (3/4)
 0.0000000000000E+00-2.0595187670703E-01
 2.0595187670703E-01
       ... hamiltonian-3 [FAILED]
  Message: Floating point value missmatch
  Starting hamiltonian-4 ... (4/4)
[...]

PR #183, opened 2024-05-24
Second version of the CEH model

thfroitzheim:
The updated second version of the CEH model was implemented. This includes the following changes: 

- Support for f-block elements up to Z=103 with explicit f-functions
- Approximate electrostatics correction of second and onsite third order based on the Mataga-Nishimoto Kernel and approximate charges (based on electronegativity-weighted CN and the total charge). The ES is atom-resolved (added a shell_resolved boolean also in the second order ES). 
- A slightly reduced electronic temperature for stability and an updated parametrization. 

Additionally, a few technical changes were made: 

- Correct ordering of the d-functions in the inline implementation of the d-transformation. (from [0, -1, 1, -2, 2] to the proper [-2, -1, 0, 1, 2]). This addresses issue #104 (may require some adaptation in the ptb implementation in xtb  @marcelmbn)
- Restructured Hamiltonian construction, such that the diatomic frame transformation, including setup of the trafo matrix, is only done once per atom pair. This is especially important for the gradient of the diatomic frame transformation as it contains many (small) matrix multiplications. 
- Compatibility between diatomic frame scaling and up to g-functions, where only d-functions are transformed.


thfroitzheim:
As far as I can tell, installing gcc via homebrew static links the wrong gfortran 14.1.0 version independent from the specified GCC_V. I now remove this static link before linking the right version (maybe there is a better way of doing this?). The gfortran-9 seems not to work on MacOS in general.

config/cmake/Findmstore.cmake
```diff
@@ -17,6 +17,7 @@
 set(_lib "mstore")
 set(_pkg "MSTORE")
 set(_url "https://github.com/grimme-lab/mstore")
+set(_rev "HEAD")
```
awvwgk: Can you please create a new release for mstore first?

src/tblite/integral/dipole.f90
```diff
@@ -379,26 +380,22 @@ pure subroutine dipole_cgto_diat_scal(cgtoj, cgtoi, r2, vec, intcut, &
...
-   !> (number of AOs that appear before the current angular momentum)
-   integer, parameter :: offset_nao(8) = [0, 1, 4, 9, 16, 25, 36, 49]
 
    integer :: ip, jp, mli, mlj, l
    real(wp) :: eab, oab, est, s1d(0:maxl2), rpi(3), rpj(3), cc, val, dip(3), pre
    real(wp) :: s3d(mlao(cgtoj%ang), mlao(cgtoi%ang))
    real(wp) :: d3d(3, mlao(cgtoj%ang), mlao(cgtoi%ang))
 
+   !> Block overlap matrix as a technical intermediate for the diatomic frame
+   real(wp) :: block_overlap(smap(max(cgtoj%ang,cgtoi%ang)+1),smap(max(cgtoj%ang,cgtoi%ang)+1))
```
awvwgk: Can we avoid this involved stack allocation?

thfroitzheim: I try to circumvent this by moving the diatomic frame transformation to the H0 in src/tblite/ceh/h0.f90. However, this requires a split in the H0 construction, where first the overlap for each pair of atoms is calculated and stored before evaluating the Hamiltonian. Long-term this would also be the case in the GP3. 
If you are fine with this split, the the overlap_cgto_diat and dipole_cgto_diat routines will not be necessary anymore (currently they are only used in the tests but I can change that).

thfroitzheim: I restructured the diatomic frame transformation now in such a way that also asymmetric pairs (i.e., between elements with different maximum angular momentum) can be efficiently treated. This removes also this complicated allocation.

thfroitzheim:
I addressed the few comments you had and also updated a few parameters for some rare/heavy elements (this has only a negligible effect on the final charges).

awvwgk:
I am maintaining this library on a volunteer basis. While I will prioritize reviewing this patch as soon as possible, given the large and extensive change I cannot promise finding time for this soon.

If you want to help me speed up this review, it would be much appreciated to have separated patches for unrelated changes, like the update of the coordination number, the change for the cartesian to spherical transformation, the update for the integrals and so forth, as many of them seem to be straight-forward to review and quicker to merge individually.

thfroitzheim:
I separated all contained changes in #185. After it is merged I will open a new PR with the remaining changes. Regrettably, further separation is not really possible since they all depend on the changed d-ordering in their tests. Separately adding the d-orbital re-ordering breaks a lot of tests and I don't have a reliable reference to update them inbetween.

PR #185, opened 2024-07-04
Preparation for Second version of the CEH model

thfroitzheim:
I separated all changes for the next CEH version (see #183), which are independent and easy to review. These include: 
- Update of the Github workflow for MacOS
- Changes in the Coordination number
- The addition to the JSON output
- Changes to the Coulomb interaction 
The fix of the integral d-order is not separated here since the diatomic frame transformation depends on the d-ordering, which would require many changes in the tests without a reliable reference.


thfroitzheim:
Some of the testing comes with the actual CEH update, where the things are used.

awvwgk (approved):
Looks good to me.

PR #186, opened 2024-07-06
Fix for integral transformation and diatomic frame overlap

thfroitzheim:
The second batch of changes for the second CEH version now with all changes to the integrals:
- fix of the d-orbital ordering from [0, -1, 1, -2, 2] to the proper [-2, -1, 0, 1, 2] (see #104)
- fix of the diatomic frame overlap gradient to include the position dependence of the diatomic axis. 
- improved Hamiltonian construction with the diatomic frame transformation restricted to atom pairs instead of CGTO pairs. 
- compatibility between diatomic frame scaling and up to g-functions, where only d-functions are transformed.
- update also the mstore version for meson
After this is merged only a small CEH update remains containing mainly new parameters and tests.


awvwgk (approved):
Looks good to me. We can go ahead and merge this

PR #187, opened 2024-07-08
Fix wheel workflow

marvinfriede:


awvwgk:
In the long term we need to migrate this workflow to cibuildwheel, like done in https://github.com/dftd3/simple-dftd3/pull/65, using the conda-forge images might be what is currently causing compatibility issues with other libraries like pytorch on pypi.

PR #188, opened 2024-07-30
Raise error for too heavy elements

marcelmbn:
## Adding error raise for too heavy elements

In the existing implementation, using actinoids (or in general elements `Z>=87`) leads to memory errors or stopped execution of the program. To prevent this (not intended behavior), with this change, each calculator checks for `any(mol%num > max_elem)`.

- changed tests
- changed calculator calls
- added error raise for each calculator (`ceh`, `gfn2`, `gfn1`, `ipea1`)


src/tblite/api/calculator.f90
```diff
@@ -99,8 +100,12 @@ function new_gfn2_calculator_api(vctx, vmol) result(vcalc) &
    call c_f_pointer(vmol, mol)
 
    allocate(calc)
-   call new_gfn2_calculator(calc%ptr, mol%ptr)
-   vcalc = c_loc(calc)
+   call new_gfn2_calculator(calc%ptr, mol%ptr, error)
+   if (allocated(error)) then
```
awvwgk, line 104: This error should be pushed to the context object, such that API users can get a clear understanding of the reason for the error

src/tblite/ceh/ceh.f90
```diff
@@ -438,10 +438,21 @@ module tblite_ceh_ceh
...
       type(structure_type), intent(in)  :: mol
+      !> Error handling
+      type(error_type), allocatable, intent(out) :: error
+      !> String out of max_elem
+      character(len=3) :: max_elem_str
+
+      write(max_elem_str, '(I3)') max_elem
+      !> Check if all atoms of mol%nat are supported (Z <= 86)
+      if (any(mol%num > max_elem)) then
+         call fatal_error(error, "No support for elements with Z >" // max_elem_str // ".")
```
awvwgk: We have a format function to avoid using internal writes in tblite

test/unit/test_ceh.f90
```diff
@@ -338,7 +338,7 @@ subroutine test_q_gen(error, mol, ref)
       integer :: i
       allocate(cn(mol%nat))
 
-      call new_ceh_calculator(calc, mol)
+      call new_ceh_calculator(calc, mol, error)
```
awvwgk, line 341: Please also handle the errors here and subsequent cases

awvwgk (changes requested):

test/unit/test_ceh.f90
marcelmbn: Do you mean something like this:
```
      call new_ceh_calculator(calc, mol, error)
      if (allocated(error)) return
```

The error itself is propagated to the test suite itself automatically, if I understand the `tests` correctly.

src/tblite/ceh/ceh.f90
```diff
@@ -438,10 +438,18 @@ module tblite_ceh_ceh
...
+   subroutine new_ceh_calculator(calc, mol, error)
       !> Instance of the CEH evaluator
       type(xtb_calculator), intent(out) :: calc
       type(structure_type), intent(in)  :: mol
+      !> Error handling
+      type(error_type), allocatable, intent(out) :: error
+
+      !> Check if all atoms of mol%nat are supported (Z <= 86)
+      if (any(mol%num > max_elem)) then
+         call fatal_error(error, "No support for elements with Z >" // format_string(max_elem, '(i3)') // ".")
```
awvwgk: To avoid leading whitespace 
```suggestion
         call fatal_error(error, "No support for elements with Z >" // format_string(max_elem, '(i0)') // ".")
```

src/tblite/api/calculator.f90
```diff
@@ -124,8 +132,12 @@ function new_ipea1_calculator_api(vctx, vmol) result(vcalc) &
    call c_f_pointer(vmol, mol)
 
    allocate(calc)
-   call new_ipea1_calculator(calc%ptr, mol%ptr)
-   vcalc = c_loc(calc)
+   call new_ipea1_calculator(calc%ptr, mol%ptr, error)
+   if (allocated(error)) then
```
awvwgk, line 136: Same for the other API calls as well ;)

src/tblite/xtb/gfn1.f90
```diff
@@ -515,11 +516,19 @@ module tblite_xtb_gfn1
...
-subroutine new_gfn1_calculator(calc, mol)
+subroutine new_gfn1_calculator(calc, mol, error)
    !> Instance of the xTB evaluator
    type(xtb_calculator), intent(out) :: calc
    !> Molecular structure data
    type(structure_type), intent(in) :: mol
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
+
+   !> Check if all atoms of mol%nat are supported (Z <= 86)
```
awvwgk: Don't use docstring in the procedure body, they lead to artifacts for the generated docs
```suggestion
   ! Check if all atoms of mol%nat are supported (Z <= 86)
```

awvwgk (changes requested):

src/tblite/xtb/gfn1.f90
marcelmbn: Okay, good to know!

src/tblite/api/calculator.f90
marcelmbn: Sure, just noticed that I have pushed it only for the example one.

app/driver_run.f90
```diff
@@ -156,13 +156,14 @@ subroutine run_main(config, error)
    call new_wavefunction(wfn, mol%nat, calc%bas%nsh, calc%bas%nao, nspin, config%etemp * kt)
 
    if (config%guess == "ceh") then
-      call new_ceh_calculator(calc_ceh, mol)
+      call new_ceh_calculator(calc_ceh, mol, error)
```
awvwgk, line 159:
```suggestion
      call new_ceh_calculator(calc_ceh, mol, error)
      if (allocated(error)) return
```

app/driver_run.f90
```diff
@@ -156,13 +156,14 @@ subroutine run_main(config, error)
...
       call new_wavefunction(wfn_ceh, mol%nat, calc_ceh%bas%nsh, calc_ceh%bas%nao, 1, config%etemp_guess * kt)
       if (config%grad) then
          call ctx%message("WARNING: CEH gradient not yet implemented. Stopping.")
          return
       end if
    end if
+   if (allocated(error)) return
```
awvwgk:
```suggestion
```

awvwgk (commented):
Errors need to be handled as soon as possible, there is no guarantee that the next calls will succeed if the previous fails, especially if they depend on the state of the calculator

app/driver_guess.f90
```diff
@@ -115,7 +115,7 @@ subroutine guess_main(config, error)
       method = "ceh"
       if (allocated(config%method)) method = config%method
       if (method == "ceh") then
-         call new_ceh_calculator(calc_ceh, mol)
+         call new_ceh_calculator(calc_ceh, mol, error)
```
awvwgk, line 118:
```suggestion
         call new_ceh_calculator(calc_ceh, mol, error)
         if (allocated(error)) return
```

awvwgk (approved):
Thanks for sharing. Looks good to me now

PR #189, opened 2024-07-30
Refactor `driver_guess`

marcelmbn:
Current implementation uses `new_gfn2_calculator` and a respective wavefunction to project guess charges on to that. 
As this is in general only a poor workaround, it prevents the user also from using guess charges with `Z>maxZ(xtb)`, e.g., using EEQ for obtaining actinoid charges.

Changes:

- General charge and atomic dipole moment array (`spin=1`) for `driver_guess` instead of the `gfn2_calculator`
- Renaming `ceh_guess` to `ceh_singlepoint` to be consistent with `gfn1, ...`
- Interface for `EEQ` and `SAD` guesses so that they can be used with partitioned shell charges or for raw atomic charges


src/tblite/wavefunction/guess.f90
```diff
@@ -29,28 +29,51 @@ module tblite_wavefunction_guess
...
+   end interface eeq_guess
 
 contains
 
-subroutine sad_guess(mol, calc, wfn)
+subroutine sad_guess_qat(mol, charges, dpat)
+   type(structure_type), intent(in) :: mol
+   real(wp), intent(inout) :: charges(:), dpat(:, :)
+
+   dpat = 0.0_wp
```
awvwgk:
```suggestion
   dpat(:, :) = 0.0_wp
```

src/tblite/wavefunction/guess.f90
```diff
@@ -29,28 +29,51 @@ module tblite_wavefunction_guess
...
+   call sad_guess_qat(mol, wfn%qat(:, 1), wfn%dpat(:, :, 1))
    call shell_partition(mol, calc, wfn)
-end subroutine sad_guess
+end subroutine sad_guess_qsh
+
+subroutine eeq_guess_qat(mol, charges, dpat)
+   type(structure_type), intent(in) :: mol
+   real(wp), intent(inout) :: charges(:), dpat(:, :)
+
+   dpat = 0.0_wp
```
awvwgk:
```suggestion
   dpat(:, :) = 0.0_wp
```

awvwgk (approved):
Looks good to me

PR #190, opened 2024-07-31
Electrostatics and second version parameters for CEH

thfroitzheim:
This is the last part of the CEH update to the second version (formerly in #183). Changes include: 
- Support for f-block elements up to Z=103 with explicit f-functions
- Approximate atom-resolved electrostatics of 2nd and onsite 3rd order with approximate charges (based on electronegativity-weighted CN and the total charge).
- A slightly reduced electronic temperature for stability and an updated parametrization.


test/unit/test_ncoord.f90
```diff
@@ -753,12 +753,12 @@ subroutine test_cn_mb01_erf(error)
...
-      & +1.46154464565793E+00_wp, +1.68476840890299E+00_wp, +3.36480649393304E+00_wp, &
-      & +1.01246571460105E+00_wp, +9.41673850989885E-01_wp, +1.86611893141095E+00_wp, &
-      & +3.82207104984211E+00_wp, +3.43775046556670E+00_wp, +3.43498366851429E+00_wp, &
-      & +5.23834442955556E+00_wp]
+      & +3.9658660209432872E+00_wp, +8.0394883202556611E-01_wp, +1.8253782118445026E+00_wp, &
+      & +1.2928270566635960E+00_wp, +1.0069012983391610E+00_wp, +9.2124160508748754E-01_wp, &
+      & +1.4316264125007339E+00_wp, +1.7548994532570443E+00_wp, +3.3916052318800585E+00_wp, &
+      & +1.0126866501857672E+00_wp, +9.3215233418135057E-01_wp, +1.7800350456604632E+00_wp, &
+      & +3.9645114512025614E+00_wp, +3.5102202229341262E+00_wp, +3.5008775101018879E+00_wp, &
+      & +5.2291536532971481E+00_wp]
```
awvwgk: I believe it would be better to not revise those examples but set the kcn value to the one the reference values were generated with.

awvwgk (changes requested):

test/unit/test_ncoord.f90
thfroitzheim: You are right. This directly tests also the custom kcn values.

awvwgk (approved):
Looks good to me

Issue #191, opened 2024-07-31
CEH single point driver uses both context and error handle


awvwgk:
The CEH single point driver uses both context and error handle, instead only one, probably the context should be passed.

ping Marcel @marcelmbn, Thomas @thfroitzheim

marcelmbn:
@thfroitzheim and I will solve this problem when implementing the CEH model update to `v2`. Thanks for the hint!

PR #192, opened 2024-07-31
Remove error container from ceh_singlepoint. Add consistent timing.

thfroitzheim:
I removed the error container from the ceh_singlepoint as suggested in #191 and added error handling in the tests (not sure but shouldn't this also be the case in the xtb tests?). 

Additionally, I restructured the timing of the ceh_singlepoint to be consistent with the xtb_singlepoint, just with one print_level higher to not overcrowd the output if CEH is used as a guess.

### DIFF
--- a/src/tblite/ceh/singlepoint.f90
+++ b/src/tblite/ceh/singlepoint.f90
@@ -75,21 +75,21 @@ contains
       !> Verbosity level of output
       integer, intent(in), optional :: verbosity
 
-      !> Molecular dipole moment
+      ! Molecular dipole moment
       real(wp) :: dipole(3)
-      !> Integral container
+      ! Integral container
       type(integral_type) :: ints
-      !> Electronic solver
+      ! Electronic solver
       class(solver_type), allocatable :: solver
-      !> Adjacency list
+      ! Adjacency list
       type(adjacency_list) :: list
-      !> Potential type
+      ! Potential type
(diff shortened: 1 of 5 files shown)


src/tblite/ceh/singlepoint.f90
```diff
@@ -91,10 +89,10 @@ subroutine ceh_singlepoint(ctx, calc, mol, error, wfn, accuracy, verbosity)
       type(container_cache) :: icache, ccache
       !> Timer
       type(timer_type) :: timer
-      real(wp) :: ttime
-
+      !> Error container
```
awvwgk: Don't use docstring markers for local variables.
```suggestion
      ! Error container
```

test/unit/test_ceh.f90
```diff
@@ -1335,7 +1350,12 @@ subroutine test_d_field_mb04(error)
       call calc%push_back(cont)
 
       ctx%verbosity = 0
-      call ceh_singlepoint(ctx, calc, mol, error, wfn, accuracy)
+      call ceh_singlepoint(ctx, calc, mol, wfn, accuracy)
+      if (ctx%failed()) then
+         call ctx%get_error(error)
+         call test_failed(error, "Singlepoint calculation failed")
```
awvwgk: This will drop the error you just retrieved from the context

awvwgk (approved):

PR #193, opened 2024-08-01
Gradient of the electrostatics potential

thfroitzheim:
I added position and lattice vector gradients of the potential for second-order and onsite third-order electrostatics. This is the first step towards gradients of the CEH charges, as they require the derivative of the Hamiltonian for the coupled-perturbed equations (a later PR). Changes include: 
- add get_potential_gradient to the potential_type and all derived types. 
- add charge derivatives to the wavefunction type. 
- add derivatives to the effective charges for the CEH electrostatics
- add explicit testing for the onsite thirdorder (also a useful starting point in the future for full thirdorder ES).


src/tblite/ceh/ceh.f90
```diff
@@ -1272,29 +1271,48 @@ end subroutine get_hubbard_derivs
...
-         do iat = 1, size(qat, 1)
-            isp = mol%id(iat)
-            izp = mol%num(isp)
-
-            qat(iat, ispin) = p_ceh_en_to_q(izp)*cn_en(iat) &
-            & + p_ceh_total_to_q * mol%charge/real(mol%nat, wp)
+      if (present(dqatdr)) dqatdr(:,:,:,:) = 0.0_wp
+      if (present(dqatdL)) dqatdL(:,:,:,:) = 0.0_wp
+
+      if (present(dqatdr) .and. present(dqatdL)) then
```
awvwgk: This branch does not check for the presence of dcn_endr and access them even if they are not provided

src/tblite/ceh/ceh.f90
```diff
@@ -1272,29 +1271,48 @@ end subroutine get_hubbard_derivs
...
       !> Molecular structure data
       type(structure_type), intent(in) :: mol
-      !> Basis set information
-      type(basis_type), intent(in) :: bas
       !> Electronegativity weighted CN, shape: [nat]
       real(wp), intent(in) :: cn_en(:)
       !> Effective atomic charges, shape: [nat, spin]
       real(wp), intent(out) :: qat(:, :)
+      !> Position and lattice vector gradient of cn_en, shape: [3, nat/3, nat]
+      real(wp), intent(in), optional :: dcn_endr(:, :, :), dcn_endL(:, :, :)
```
awvwgk: Please split those over multiple lines

src/tblite/coulomb/charge/type.f90
```diff
@@ -183,6 +185,56 @@ subroutine get_potential(self, mol, cache, wfn, pot)
...
+      ! Off-diagonal Coulomb matrix derivative
+      pot%dvatdr(:, :, :, 1) = dadr
+
+      do ic = 1, 3 
+         do iat = 1, mol%nat
+            ! Diagonal Coulomb matrix derivative
+            pot%dvatdr(ic, iat, iat, 1) = - sum(dadr(ic, :, iat))
+
+            ! Charge derivative
+            call symv(ptr%amat, wfn%dqatdr(ic, iat, :, 1), pot%dvatdr(ic, iat, :, 1), beta=1.0_wp)
```
awvwgk: Please explicitly copy those slices into a temporary array to avoid that the compiler creates them in an inefficient way

src/tblite/coulomb/charge/type.f90
```diff
@@ -183,6 +185,56 @@ subroutine get_potential(self, mol, cache, wfn, pot)
...
+         end do
+         
+         ! Coulomb matrix derivative 
+         pot%dvatdL(ic, :, :, 1) = pot%dvatdL(ic, :, :, 1) + dadL(ic, :, :)
+         do jc = 1, 3
+            ! Charge derivative
+            call symv(ptr%amat, wfn%dqatdL(ic, jc, :, 1), pot%dvatdL(ic, jc, :, 1), beta=1.0_wp)
```
awvwgk: Same here, please don't rely on the compiler to create temporary contiguous arrays for you

src/tblite/coulomb/multipole.f90
```diff
@@ -267,6 +269,24 @@ subroutine get_potential(self, mol, cache, wfn, pot)
...
+   class(damped_multipole), intent(in) :: self
+   !> Molecular structure data
+   type(structure_type), intent(in) :: mol
+   !> Wavefunction data
+   type(wavefunction_type), intent(in) :: wfn
+   !> Density dependent potential
+   type(potential_type), intent(inout) :: pot
+   !> Reusable data container
+   type(container_cache), intent(inout) :: cache
+   type(coulomb_cache), pointer :: ptr
```
awvwgk: Unused

src/tblite/coulomb/thirdorder.f90
```diff
@@ -170,6 +172,33 @@ subroutine get_potential(self, mol, cache, wfn, pot)
...
+   !> Reusable data container
+   type(container_cache), intent(inout) :: cache
+   !> Wavefunction data
+   type(wavefunction_type), intent(in) :: wfn
+   !> Density dependent potential
+   type(potential_type), intent(inout) :: pot
+
+   integer :: iat, izp, ii, ish
+
+   if (.not. self%shell_resolved) then
```
awvwgk: What happens in the shell resolved case? Can we at least raise an error if this is not supported rather than falling through silently

awvwgk (changes requested):

src/tblite/coulomb/thirdorder.f90
thfroitzheim: You are right that it doesn't make much sense to break the support for both shell- and atom-resolved ES over this small addition. I added the shell-resolved case and tested it in the CEH context with atomic charges distributed over the shells.

awvwgk (approved):
Looks good to me

PR #196, opened 2024-08-07
Update CI to latest state

marcelmbn:
- remove `macOS` Intel compiler support
- use existing compilers rather than installing them from scratch for `macOS`
- update GitHub actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,10 +18,6 @@ env:
     ase
     matplotlib==3.4.2
     qcelemental
-  MACOS_BASEKIT_URL: >-
-    https://registrationcenter-download.intel.com/akdlm/irc_nas/17969/m_BaseKit_p_2021.3.0.3043.dmg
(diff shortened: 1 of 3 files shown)


marcelmbn:
The [Linux image used for building the wheel](https://hub.docker.com/r/condaforge/linux-anvil-cos7-x86_64) was probably updated to `CentOS7` recently.  Thus, the wheel building process triggered on `push` fails as `auditwheel` cannot repair the wheel to be compatible with `glibc-2.12`, which is, in my view, not necessary as this corresponds to, e.g., Ubuntu versions `<14.04`.
Furthermore, I've updated all the GitHub actions versions.

marcelmbn:
Good reads on compatible systems and the correlation with respect to `manylinux` compatibility can be found here: https://github.com/mayeut/pep600_compliance
https://peps.python.org/pep-0600/

awvwgk (approved):

PR #199, opened 2024-08-17
Add py.typed file to indicate support for type checkers

Andrew-S-Rosen:
As recommended in [pep 0561](https://peps.python.org/pep-0561/), a blank `py.typed` marker should be included when type hints are used so downstream codes can type check with `mypy` and similar tools.


awvwgk (changes requested):
The file needs to be included in the distribution, please add it to the meson.build file.

Andrew-S-Rosen:
@awvwgk: Thanks for pointing that out. I added it to `pyproject.toml` just now but am not familiar with the `meson.build` format. I will need some assistance on that.

python/pyproject.toml
```diff
@@ -32,3 +32,6 @@ optional-dependencies.test = [
   "pytest",
   "pytest-cov",
 ]
+
+[tool.setuptools.package-data]
```
awvwgk: We don't use setuptools for this project. The package is defined in `python/tblite/mesin.build` and `py.typed` can be added next to use regular Python sources there.

Andrew-S-Rosen: I have removed the lines from `pyproject.toml` but remain unsure how `meson.build` works. I made a change but don't know if I did it right.

awvwgk (approved):
Thanks for sharing. I added the py.typed file to the list of installed files.

python/pyproject.toml
awvwgk: GH mobile makes it hard to edit PRs, finally got to open my laptop for this. Generally, meson is quite transparent about building and installing projects, so there are no implicit rules or automatic discovery steps like in setuptools. While setuptools is the defacto standard for building Python, it makes it incredible hard to build non-Python components. In meson both Python and non-Python have first class support, which makes it easier to handle the compilation at the cost of slightly more verbose installation instructions for the Python part.

Andrew-S-Rosen:
Thank you for your assistance.

PR #204, opened 2024-09-20
Fix real precisions

### LINKED ISSUE #91: Compiling with ubuntu-latest, meson, intel, 2021.2.0 fails
Currently, the ifort CI fails. This occurred first in #90

foxtran:
Fixes #91

The problem of 91 comes from usage of `-r8` flag in xtb builds, while tblite does not have this flag in CI and therefore, can not detect such kind of bugs. The same issue can be observed with GCC via passing `-fdefault-real-8` flag during compilation. The flag that helps to debug such kind of issue for GCC: `-Wconversion`. Unfortunately, it does not help to detect these bugs in `parameter` definitions.


faust403:
It would still be better to rewrite everything on rust tbh

awvwgk (approved):
Good catch. Thanks for sharing

foxtran:
@awvwgk, just trying to compile xtb with ifx :(

Check other repos too :D

PR #209, opened 2024-10-26
Update subproject versions in meson, cmake and fpm (simple-dftd3, dftd4, ...)

foxtran:
See https://github.com/grimme-lab/xtb/issues/1120


thfroitzheim:
If we are already updating the dependencies, we should do so also for the cmake and fpm builds (ideally again to the same tagged version as in meson). At least cmake is also relevant for the xtb build and both still point to the old simple-dftd3 repository.

foxtran:
Oh... It has different version of dependencies for different build systems... What a mess...

awvwgk (approved):

PR #215, opened 2024-11-08
Pin dependency versions also in meson.build

thfroitzheim:
I also pinned the dependency versions in the meson.build files. We might want to use a ">=" but to allow for a xtb build where all dependencies are consistent, this seems to require specification here to.

### DIFF
--- a/config/meson.build
+++ b/config/meson.build
@@ -123,7 +123,7 @@ lib_deps += mctc_dep
 # Create DFT-D4 library as subproject
 dftd4_dep = dependency(
   'dftd4',
-  version: '>=3.4.0 <4.0.0',
+  version: ['>=3.4.0', '<4'],
   fallback: ['dftd4', 'dftd4_dep'],
   default_options: ['default_library=static', 'api=false', 'python=false'],
 )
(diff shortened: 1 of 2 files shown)


awvwgk (changes requested):
Please retain the version ranges. If there is no breaking API change from the dependencies are a new API that requires a certain lower bound we should leave the flexibility to discover all compatible versions of a dependency.

config/meson.build
```diff
@@ -123,7 +123,7 @@ lib_deps += mctc_dep
 # Create DFT-D4 library as subproject
 dftd4_dep = dependency(
   'dftd4',
-  version: '>=3.0.0',
+  version: '>=3.4.0',
```
awvwgk: This needs an upper bound of v4

awvwgk (changes requested):

config/meson.build
```diff
@@ -123,7 +123,7 @@ lib_deps += mctc_dep
 # Create DFT-D4 library as subproject
 dftd4_dep = dependency(
   'dftd4',
-  version: '>=3.0.0',
+  version: '>=3.4.0 <4.0.0',
```
marvinfriede:
```suggestion
  version: ['>=3.4.0', '<4'],
```

awvwgk (approved):

PR #221, opened 2024-11-19
fix: prevent calculator.get from printing properties

apaulTQ:
When getting properties from the calculator I got prints to stdout of the property arrays. I removed the print statements from get_post_processing_dict to fix this. Let me know if they should rather be hidden by verbosity


awvwgk (approved):
Good catch, thanks for fixing.

apaulTQ:
Thanks for merging. Is  there an approximate date until which one can expect the release of bugfixes for tblite 0.4?

PR #222, opened 2024-11-27
Atomic charges (`qat`) in `tblite.json` dump

marcelmbn:
### What I want to have

- atomic charges (coming from `wfn%qat(:,1)`) in the `tblite.json` dump, accessible via `--json`

## "Naive" implementation

- see commit 

## "Right way" of doing it

... could probably also be via `add_post_processing` or similar. @awvwgk @pitsteinbach What do you think about it? 

Tbh, (even though that's a bit off-topic) I don't really understand how to access these post-processing results, e.g., Wiberg-Mayer bond orders via CLI as a `stdout` or via file dump. 😄 IMHO, they should be exported to the `JSON` file as well.

### DIFF
--- a/app/driver_run.f90
+++ b/app/driver_run.f90
@@ -271,7 +271,7 @@ subroutine run_main(config, error)
       call ascii_levels(ctx%unit, config%verbosity, wfn%homo, wfn%emo, wfn%focc, 7)
       call post_proc%dict%get_entry("molecular-dipole", dpmom)
       call post_proc%dict%get_entry("molecular-quadrupole", qpmom)
-      
+
       call ascii_dipole_moments(ctx%unit, 1, mol, wfn%dpat(:, :, 1), dpmom)
       call ascii_quadrupole_moments(ctx%unit, 1, mol, wfn%qpat(:, :, 1), qpmom)
    end if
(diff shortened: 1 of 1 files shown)


pitsteinbach:
The post processing results are stored in a dictionary, which itself can be easily dumped to a toml file. That's all prepared. I have even written some tests to do exactly that, however I never connected it to CLI. But we can also include a dump to JSON and then use either `--json` or `--toml` to dump to a file. That would be my suggestion. Anyway thanks for letting me know.

pitsteinbach:
About your feature request and partially also the toml dump, the `xtbml` features that will be published soon collect among other the atomic charges and dump them as a toml. You can have a look at my `tblite` fork and the `xtbml-new-form` branch. There when you enter  `--post-processing xtbml`, all features are computed and then dumped to `post_processing.toml`.
So I would propose that I will (re)open a PR for the `xtbml` features, then we can add maybe also a `json` dump and that would solve both problems? @awvwgk what's your opinion on that?

marcelmbn:
> About your feature request and partially also the toml dump, the `xtbml` features that will be published soon collect among other the atomic charges and dump them as a toml. You can have a look at my `tblite` fork and the `xtbml-new-form` branch. There when you enter `--post-processing xtbml`, all features are computed and then dumped to `post_processing.toml`. So I would propose that I will (re)open a PR for the `xtbml` features, then we can add maybe also a `json` dump and that would solve both problems? @awvwgk what's your opinion on that?

Sounds good to me!

pitsteinbach:
@marcelmbn the `xtbml` PR is merged so you can get the atomic charges in a `toml` file or via the Python API.

marcelmbn:
> @marcelmbn the `xtbml` PR is merged so you can get the atomic charges in a `toml` file or via the Python API.

Great, thanks for the update! Then we can close here, I guess.

PR #225, opened 2024-12-05
Add TBLiteValueError to dict() method in Results Container python

pitsteinbach:
Currently part of the results throw a `ValueError` or `TBLiteValuError`, whereas other throw a RunTime Error if they are not contained in the results container. Now all the results coming from the dictionary use the `TBLiteValueError`, I also added it to the `Results.dict()` method as it returned an exception if a post-processing value is added manually.
I also added a check for this situation to `test_interface.py` and changed the error checking to the `TBLiteValueError` type.

### DIFF
--- a/python/tblite/interface.py
+++ b/python/tblite/interface.py
@@ -25,7 +25,7 @@ from typing import Any, List, Optional, Union
 import numpy as np
 
 from . import library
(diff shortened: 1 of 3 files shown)


awvwgk (approved):
Looks good to me

PR #155, opened 2024-01-26
Updated spGFN2 spin constants

benbaed:
The constants were now computed using a more sophisticated script (minor changes in the last digits might occur)
The rare gases were fixed (now actually PBE is used, before it was HF by accident)
Cr, Cu, Y, Zr, Pd, Pt & Au were changed in the electronic configuration to avoid "jumps", the d-related constants changed
B, Cl, Ir & Bi were adjusted in a minor way.

PR #228, opened 2025-02-13
Avoid division by zero at fermi.f90

foxtran:
Noticed for `tblite/post-processing` test.

Stacktrace:
```
Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x7fb83d606b4f in ???
#1  0x9002a5 in get_fermi_filling_
	at /subprojects/tblite/src/tblite/wavefunction/fermi.f90:96
#2  0x900de5 in __tblite_wavefunction_fermi_MOD_get_fermi_filling
	at /subprojects/tblite/src/tblite/wavefunction/fermi.f90:46
#3  0x850c79 in __tblite_scf_iterator_MOD_get_density
	at /subprojects/tblite/src/tblite/scf/iterator.f90:318
#4  0x856eb1 in __tblite_scf_iterator_MOD_next_scf
	at /subprojects/tblite/src/tblite/scf/iterator.f90:110
```


marvinfriede:
CI was already fixed in #155.

foxtran:
So I do not need to do it like in xtb :D

src/tblite/wavefunction/fermi.f90
```diff
@@ -93,7 +93,11 @@ subroutine get_fermi_filling_(homo, kt, emo, occ, e_fermi)
          total_number = total_number + fermifunct
          total_dfermi = total_dfermi + dfermifunct
       end do
-      change_fermi = (occt-total_number)/total_dfermi
+      if (total_dfermi > 1e-80_wp) then
```
awvwgk: Have you considered using `epsilon(1.0_wp)` here instead of a fixed constant?

foxtran: 1e-16 is quite big, to be honest. Probably, something like `sqrt(tiny(1.0_wp))` should be better (it is like 1e-154). And `1e-80` is something like `tiny(1.0_wp)**(1./4.)` =)

thfroitzheim: Then let's replace it with `sqrt(tiny(1.0_wp))` to be specific for the precision. You could add another parameter at the top.

marvinfriede (approved):

Issue #71, opened 2022-06-21
More general solver object for returning density matrix


awvwgk:
Currently the solver object only returns the eigenvalues and coefficient matrices, for solvers working without diagonalization we usually only return the density matrix.

- [ ] handle electronic filling (finite temperature)
- [ ] collect `get_density` routine in solver object

https://github.com/tblite/tblite/blob/a0affbf359a6b719a06f27a3ab65d1764dc8d6de/src/tblite/scf/iterator.f90#L292-L336

cc @cbannwarth

awvwgk:
Available with https://github.com/tblite/tblite/pull/271

PR #237, opened 2025-03-17
Move coordination number and common data to mctc-lib and add DFT-D4S

thfroitzheim:
PR to move the coordination number to the mctc-lib (see PR [#71](https://github.com/grimme-lab/mctc-lib/pull/71) there). This avoids duplicated code in all our projects and makes the addition of new coordination numbers simpler. The full functionality and all tests were moved to mctc-lib. Additionally, the common data (electronegativity and radii) were moved to mctc-lib. 

This PR starts as a draft pointing to the mctc-lib [branch](https://github.com/thfroitzheim/mctc-lib/tree/ncoord) until it is merged.


thfroitzheim:
I had to update the version of DFT-D4 used in tblite to be consistent with the updated multicharge (otherwise we would use two different versions). With this, I also introduced the new D4S feature and integrated it into the parameter toml file (so far, no method uses it as default). I also added tests for the dispersion, specifically the potential, which was never tested (also not in DFT-D4).

src/tblite/param/dispersion.f90
```diff
@@ -29,7 +29,7 @@ module tblite_param_dispersion
    public :: count
 
 
-   character(len=*), parameter :: k_d3 = "d3", k_d4 = "d4", k_sc = "sc", &
+   character(len=*), parameter :: k_d3 = "d3", k_d4 = "d4", k_sc = "sc", k_smooth="smooth", &
```
marvinfriede:
```suggestion
   character(len=*), parameter :: k_d3 = "d3", k_d4 = "d4", k_sc = "sc", k_smooth = "smooth", &
```

src/tblite/disp/d4.f90
```diff
@@ -180,15 +250,40 @@ subroutine get_potential(self, mol, cache, wfn, pot)
...
 
    call view(cache, ptr)
 
    call self%model%weight_references(mol, ptr%cn, wfn%qat(:, 1), ptr%gwvec, ptr%vvec, &
       & ptr%dgwdq)
 
-   call gemv(ptr%dispmat, ptr%gwvec, ptr%vvec)
-   ptr%vvec(:, :) = ptr%vvec * ptr%dgwdq
-   pot%vat(:, 1) = pot%vat(:, 1) + sum(ptr%vvec, 1)
+   if ( self%model%ncoup > 1 ) then
```
marvinfriede:
```suggestion
   if (self%model%ncoup > 1) then
```

src/tblite/disp/d4.f90
```diff
@@ -155,14 +203,36 @@ subroutine get_energy(self, mol, cache, wfn, energies)
...
+   integer :: iat, jat, izp, jzp, iref, jref
 
    call view(cache, ptr)
 
    call self%model%weight_references(mol, ptr%cn, wfn%qat(:, 1), ptr%gwvec)
 
-   call gemv(ptr%dispmat, ptr%gwvec, ptr%vvec, alpha=0.5_wp)
-   ptr%vvec(:, :) = ptr%vvec * ptr%gwvec
-   energies(:) = energies + sum(ptr%vvec, 1)
+   if ( self%model%ncoup > 1 ) then
```
marvinfriede:
```suggestion
   if (self%model%ncoup > 1) then
```

src/tblite/disp/d4.f90
```diff
@@ -62,27 +66,71 @@ module tblite_disp_d4
...
+   !> Molecular structure data
+   type(structure_type), intent(in) :: mol
+   !> Damping parameters
+   real(wp), intent(in) :: s6, s8, a1, a2, s9
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
+
+   type(d4s_model), allocatable :: tmp
+
+   self%label = label_D4S
```
marvinfriede:
```suggestion
   self%label = label_d4s
```

src/tblite/disp/d4.f90
```diff
@@ -62,27 +66,71 @@ module tblite_disp_d4
...
    !> Molecular structure data
    type(structure_type), intent(in) :: mol
+   !> Damping parameters
    real(wp), intent(in) :: s6, s8, a1, a2, s9
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
+
+   type(d4_model), allocatable :: tmp
+
+   self%label = label_D4
```
marvinfriede:
```suggestion
   self%label = label_d4
```

src/tblite/disp/d4.f90
```diff
@@ -62,27 +66,71 @@ module tblite_disp_d4
       procedure :: get_gradient
    end type d4_dispersion
 
-   character(len=*), parameter :: label = "self-consistent DFT-D4 dispersion"
+   character(len=*), parameter :: label_D4 = "self-consistent DFT-D4 dispersion"
+   character(len=*), parameter :: label_D4S = "self-consistent DFT-D4S dispersion"
```
marvinfriede:
```suggestion
   character(len=*), parameter :: label_d4 = "self-consistent DFT-D4 dispersion"
   character(len=*), parameter :: label_d4s = "self-consistent DFT-D4S dispersion"
```

src/tblite/param/hamiltonian.f90
```diff
@@ -82,18 +83,22 @@ subroutine load_from_toml(self, table, error)
...
 
-   call get_value(child, k_cn, self%cn, stat=stat)
+   ! Read string version of CN type and convert to integer
+   call get_value(child, k_cn, tmp_cn, stat=stat)
    if (stat /= 0) then
       call fatal_error(error, "Invalid entry for CN type in Hamiltonian")
       return
    end if
+   self%cn = get_cn_count_id(tmp_cn)
+   if (allocated(error)) return
```
marvinfriede:
```suggestion
```

marvinfriede: This error cannot be allocated.

src/tblite/xtb/calculator.f90
```diff
@@ -204,17 +207,19 @@ subroutine add_basis(calc, mol, param, irc)
...
    type(param_record), intent(in) :: param
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
 
-   if (allocated(param%hamiltonian%cn)) then
-      call new_ncoord(calc%ncoord, mol, cn_type=param%hamiltonian%cn)
-   end if
+   integer :: cn_count_type
+
+   call new_ncoord(calc%ncoord, mol, cn_count_type=param%hamiltonian%cn, error=error)
```
marvinfriede: Does removing the `if (allocated(param%hamiltonian%cn))` check yield the same behavior as before? I think, before, it just silently skipped creating the CN if no parameters are given. Does it now just fail? Do we want that?

src/tblite/xtb/calculator.f90
```diff
@@ -247,11 +254,18 @@ subroutine add_dispersion(calc, mol, param)
...
       if (par%d3) then
          allocate(d3)
-         call new_d3_dispersion(d3, mol, s6=par%s6, s8=par%s8, a1=par%a1, a2=par%a2, s9=par%s9)
+         call new_d3_dispersion(d3, mol, s6=par%s6, s8=par%s8, &
+            & a1=par%a1, a2=par%a2, s9=par%s9, error=error)
          call move_alloc(d3, calc%dispersion)
       else
          allocate(d4)
-         call new_d4_dispersion(d4, mol, s6=par%s6, s8=par%s8, a1=par%a1, a2=par%a2, s9=par%s9)
+         if (par%smooth) then
```
marvinfriede, line 265: Does this break custom parametrizations that do not use any dispersion? If no D3 is used, this automatically picks D4 and allocates an error if this is also not present, right? If the error is allocated, we return in the `new_xtb_calculator` subroutine without adding potentially remaining contributions.

src/tblite/xtb/gfn1.f90
```diff
@@ -923,7 +929,7 @@ subroutine export_gfn1_param(param)
       &2017, 13, 1989-2009. DOI: 10.1021/acs.jctc.7b00118"
 
    associate(par => param%hamiltonian)
-      par%cn = "exp"
+      par%cn = cn_count%exp
```
marvinfriede: This now exports the integer.

src/tblite/xtb/gfn2.f90
```diff
@@ -901,7 +910,7 @@ subroutine export_gfn2_param(param)
       &2019, 15, 1652-1671. DOI: 10.1021/acs.jctc.8b01176"
 
    associate(par => param%hamiltonian)
-      par%cn = "gfn"
+      par%cn = cn_count%dexp
```
marvinfriede: This also exports an integer now.

src/tblite/xtb/ipea1.f90
```diff
@@ -914,7 +920,7 @@ subroutine export_ipea1_param(param)
       &2017, 8, 4879. DOI: 10.1039/c7sc00601b"
 
    associate(par => param%hamiltonian)
-      par%cn = "exp"
+      par%cn = cn_count%exp
```
marvinfriede: This also exports an integer now.

test/unit/test_ceh.f90
```diff
@@ -569,14 +578,14 @@ subroutine test_scaled_selfenergy_accl6(error)
...
       real(wp), parameter :: scaled_selfenergy(nsh) = reshape([&
-      & 2.73904391613233_wp    , -4.2731759819762E-01_wp, -5.4115371441002E-01_wp, &
-      &-1.67143635735189E-02_wp, -4.3819259611367E-01_wp, -2.9082233002044E-01_wp, &
-      & 7.02743423215177E-02_wp, -4.3819425191345E-01_wp, -2.9084728314198E-01_wp, &
-      & 7.02681834728156E-02_wp, -4.3822006265352E-01_wp, -2.9084720659706E-01_wp, &
-      & 7.02683075671165E-02_wp, -4.3822174122876E-01_wp, -2.9085944552609E-01_wp, &
-      & 7.02652903250950E-02_wp, -4.3819504378374E-01_wp, -2.9083806237387E-01_wp, &
-      & 7.02704650332720E-02_wp, -4.3818833398193E-01_wp, -2.9082921173350E-01_wp, &
-      & 7.02726245699396E-02_wp], shape(scaled_selfenergy))
+      & 3.47146858519381_wp,     -4.35870279866614E-01_wp, -5.8999616375307806E-01_wp, &
```
marvinfriede: Why did this change?

src/tblite/xtb/calculator.f90
thfroitzheim: We check directly before the if statement if dispersion is allocated (see [here](https://github.com/thfroitzheim/tblite/blob/2918601af20b8ae8567bd64bed24b0620f63cffb/src/tblite/xtb/calculator.f90#L253)). If this is not the case we don't get into the selection at all. So it should be fine

test/unit/test_ceh.f90
thfroitzheim: The tabulated electronegativities have changed. In tblite they were dummy values for the actinoids, while I took the values from the D4 update for mctc-lib. For the actual CEH this makes no difference, since the EN is a parameter. Only during testing without these parameters I used the tabulated electronegativities.

src/tblite/xtb/gfn1.f90
thfroitzheim: This is an unnecessary break of the API. So I changed it back and translate the string later

src/tblite/xtb/calculator.f90
marvinfriede: Ah, I overlooked the check you linked.

awvwgk:
Thomas @thfroitzheim, what is the status on this patch?

thfroitzheim:
@awvwgk this patch requires a new release of DFT-D4 (it points to HEAD on dftd4). The new dftd4 release is API breaking, and since we are currently working on some restructuring of D4 (likely breaking the API again), we want to wait with a new dftd4 release. 
So I guess we should move this PR to the tblite 0.6.0 milestone.

marvinfriede (approved):

PR #238, opened 2025-03-24
Fix debug compilation

marvinfriede:
Compiling with meson (gcc) and debug settings breaks tests with:

```
At line 253 of file ../src/tblite/xtb/singlepoint.f90
Fortran runtime error: Allocatable actual argument 'icache' is not allocated
```

Debug flags:

```sh
meson setup build --buildtype=debug --libdir=lib -Dfortran_args="-fcheck=all -g -fbacktrace -fno-omit-frame-pointer -ffpe-trap=invalid,zero,overflow"
```


lukaswittmann (approved):
LGTM!

PR #239, opened 2025-03-24
Remove unused variables

marvinfriede:
Removes pretty much all unused variables in tblite.
Also, compile with `-Wno-maybe-uninitialized` to ignore the gcc warnings.


src/tblite/post_processing/list.f90
```diff
@@ -77,25 +77,24 @@ subroutine print_timer(self, prlevel, ctx)
...
-   class(post_processing_list),intent(in) :: self
+   class(post_processing_list), intent(in) :: self
    type(structure_type), intent(in) :: mol
    type(results_type), intent(inout) :: res
-   integer :: i, n
-   real(wp), allocatable :: tmp_array(:)
-   character(len=:), allocatable :: tmp_label
+   ! integer :: i, n
+   ! real(wp), allocatable :: tmp_array(:)
+   ! character(len=:), allocatable :: tmp_label
```
awvwgk: I guess those can be removed as well

test/unit/test_ipea1_xtb.f90
```diff
@@ -447,6 +449,17 @@ subroutine test_s_mb03(error)
    call new_wavefunction(wfn, mol%nat, calc%bas%nsh, calc%bas%nao, 1, kt)
    call xtb_singlepoint(ctx, mol, calc, wfn, acc, energy, gradient, sigma, 0)
 
+   call numdiff_grad(ctx, mol, calc, wfn, numgrad)
```
awvwgk, line 456: I don't think that part should be here. This routine test the numerical viral calculation. For a numerical gradient I would prefer a separate test

awvwgk (approved):
Thanks for doing this clean up.

test/unit/test_ipea1_xtb.f90
marvinfriede: I made a separate test. I also noticed that the virial test is commented out and actually fails. Is this a known issue?

PR #240, opened 2025-03-25
Density Matrix Purification Solvers with optional GPU acceleration

pitsteinbach:
With this PR I want to introduce DMP solvers implemented in our `GAMBITS` library to `tblite`. Additionally, a `cuSOLVER´-based  DSYGVD solver is also added. `GAMBITS` depends on LAHVA which provides a BLAS LAPACK interface with easy switching between CPU and GPU based execution.
In general we would need to discuss how our CD/CI can test for GPU builds.

There are also a few changes in this PR that affect the solver type, those are mainly needed when interfacing `tblite` to CREST or another program based on MD simulations where we would like to have a solver that can exist for multiple singlepoint calculations. Additional work, will also be needed in `CREST` but that is also prepared.

Further details about the method and it's implementation can be found here: https://chemrxiv.org/engage/chemrxiv/article-details/678e35ecfa469535b9c804c4

ToDos:

- [x] CI/CD adaptation for GPU builds
- [x] documentation on how to build for GPUs
- [x] documentation on DMP shemes


.vscode/settings.json
awvwgk, line 1: Let's not add the VS code config here.

pitsteinbach: my bad, I will clean that up once I get the wheels to compile properly.

python/meson_options.txt
```diff
@@ -20,3 +20,11 @@ option(
   value: 'python3',
   description: 'Python version to link against.',
 )
+
+option(
+  'gpu',
```
awvwgk: I don't think we need this option in the Python bindings. You can set the GPU support via `-Dtblite:gpu=...`

src/tblite/scf/iterator.f90
```diff
@@ -302,8 +304,39 @@ subroutine next_density(wfn, solver, ints, ts, error)
...
+   class(solver_type), intent(inout) :: solver
+   !> Integral container
+   type(integral_type), intent(in) :: ints
+   !> Electronic entropy
+   real(wp), intent(inout) :: ts
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
 
-   real(wp) :: e_fermi, stmp(2)
+   real(wp) :: e_fermi = 0.0_wp, stmp(2)
```
awvwgk: Please don't initialize any variable with implicit save. This is prone to create bugs.

src/tblite/scf/iterator.f90
```diff
@@ -302,8 +304,39 @@ subroutine next_density(wfn, solver, ints, ts, error)
...
+      if (solver%got_transform() .and. solver%purification_success) then
+         call solver%get_density(wfn%coeff, ints%overlap, wfn%emo, wfn%focc, wfn%density, error)
+         
+
+         if (.not.solver%purification_success) then 
+            call get_density_from_coeff(wfn, solver, ints, ts, error)
+         end if
+      else
+         call get_density_from_coeff(wfn, solver, ints, ts, error)
+      end if
```
awvwgk: Can this be implemented in the purification solver rather than the SCF iterator? It seems like an implementation detail, we don't really need to care about here.

subprojects/gambits.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory = gambits
+url = https://git.rwth-aachen.de/bannwarthlab/gambits.git
+revision = HEAD
```
awvwgk: Can we pin this to a tag?

pitsteinbach: That's the plan once everything is stable.

.github/workflows/wheel.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         git config user.name "dummy"
         git subtree add --prefix python/subprojects/tblite . HEAD
     - run: |
-        python -m build python/ --sdist --outdir . -n
+        python -m build python/ --sdist --outdir . -n --config-setting setup-args="-Dgambits:gpu=false" --config-setting setup-args="-Dlahva:gpu=false"
```
awvwgk: If this is an issue for the Python build, can we make gambits and lahva GPU an opt-in rather than opt-out? Or preferably instead of having them as boolean, please use a feature flag that disables automatically if no GPU is available.

pitsteinbach: To the first point, if the python build would be using the gpu flag from the main project where it is an opt-in than that would be way easier. But somehow the yielding mechanism of meson doesn't really kick in here.

Just because I have no experience with this python builds, are these wheels build locally upon every install?
I can just use meson to check for nvcc availabilty and then switch the GPU capability on or off based on that.

PR #241, opened 2025-03-30
ASE support for solvation 

### LINKED ISSUE #229: Use solvation with spin polarization in ase
Hello,

thank you for this great package! I wonder, is there any way to use spin polarization and solvation at the same time, preferrably in the ase calculator? According to the title of #150 it seems to me like it should be possible, since the tblite python API lists both the spin polarization and the solvent as parameters to set. However, the documentation of the ase calculator does not mention solvation at all. Is this implemented or not?

thfroitzheim:
I added support for the solvation models to ASE (see #229). It reproduces all the test cases we check for the interface itself. Maybe there is a cleaner way to handle the optional second argument for the born kernel. For QCSchema, I just updated the documentation since I am not sure how it handles additional arguments (they are also not covered in the tests, i.e., for spin polarization). 

In addition, I set thresholds also in the solvation tests, as the default 1e-6 does not distinguish between the `p16` and `still` kernels for ALPB and GBSA.


pitsteinbach (approved):
looks good to me, thanks for taking care.

python/tblite/qcschema.py
```diff
@@ -55,9 +55,12 @@
  verbosity           Set verbosity of printout            integer (1)
  electric-field      Uniform electric field               Field vector
  spin-polarization   Spin polarization                    Scaling factor
- alpb-solvation      ALPB implicit solvation              Epsilon or solvent
- cpcm-solvation      CPCM implicit solvation              Epsilon or solvent
-=================== ==================================== =====================
+ alpb-solvation      ALPB implicit solvation              Solvent name, solution state (optional)
```
marvinfriede, line 58: Does solvation work with qcschema or are we documenting something that's actually broken? (Unknown keyword [here](https://github.com/tblite/tblite/blob/07a9e662f33ec5c9dfd3f09786e973a9edca0708/python/tblite/qcschema.py#L204)?)

thfroitzheim: Yes, it actually works. I added tests for solvation and spin-polarization to make sure.

python/tblite/test_qcschema.py
```diff
@@ -181,6 +260,101 @@ def test_qcschema(atomic_input: AtomicInput, return_result: Any) -> None:
...
+    # fmt: on
+
+
+@pytest.mark.skipif(qcel is None, reason="requires qcelemental")
+def test_qcschema_solvation(atomic_input_solvation: AtomicInput, return_result_solvation: Any) -> None:
+    """Test qcschema interface."""
+    atomic_result = run_schema(atomic_input_solvation)
+
+    assert atomic_result.success
+    print(atomic_result.return_result)
```
marvinfriede:
```suggestion
```

python/tblite/test_qcschema.py
```diff
@@ -181,6 +260,101 @@ def test_qcschema(atomic_input: AtomicInput, return_result: Any) -> None:
...
 
 
+@pytest.fixture(params=[
+      {"cpcm-solvation": 7.0}, 
+      {"alpb-solvation": ["water", "bar1mol"]},
+      {"gbsa-solvation": ["methanol", "reference"]},
+      {"gbe-solvation": [7.0, "p16"]},
+      {"gb-solvation": [7.0, "still"]},
+   ])
+def solvation(request) -> str:
```
marvinfriede:
```suggestion
def solvation(request) -> dict:
```

marvinfriede (approved):

marvinfriede:
Thanks for adding the tests!

PR #242, opened 2025-03-31
Add tutorials on using Python package

awvwgk:
- [x] enable jupyter notebook for readthedocs
     - [x] change to conda environment
     - [x] build tblite extension module
- [x] singlepoint calculations
     - [x] computing energy from numpy arrays in Python
     - [x] reading xyz file to compute energy
     - [x] obtaining HOMO energy
     - [x] computing vertical ionization potential
     - [x] computing Fukui indices
     - [x] using orbitals and overlap to compute atomic populations


doc/tutorial/python/singlepoint.rst
```diff
@@ -0,0 +1,334 @@
...
+
+   Since xTB is a semiempirical method it makes some approximations which result in a strong self-interaction for a free electron.
+   This value can be computed exactly from the xTB parameters or determined empirically.
+   For a full derivation checkout Ref. :footcite:`neugebauer2020`.
+
+
+Fukui indices from partial charges
+----------------------------------
+
+While the molecular ionization potential is a great descriptor for the whole molecule, xTB also provides many properties which are atom resolved
```
marvinfriede:
```suggestion
While the molecular ionization potential is a great descriptor for the whole molecule, xTB also provides many properties, which are atom-resolved.
```

doc/tutorial/python/singlepoint.rst
```diff
@@ -0,0 +1,334 @@
...
+Computing the Fukui index provides a simple descriptor for chemical reactivity which we can compute from the partial charges according to the following equations:
```
marvinfriede:
```suggestion
Computing the Fukui index provides a simple descriptor for chemical reactivity, which we can compute from the partial charges according to the following equations: 
```

V-Alizade:
This looks great! 
Can we include the examples and tutorials as a Jupyter Notebook to make them easier to view and update?

awvwgk:
That should be possible, for this we should update our configuration for readthedocs to use a conda-environment with Fortran compiler as described [here](https://docs.readthedocs.com/platform/stable/guides/conda.html#compiling-your-project-sources)

pitsteinbach:
The part about localizing frontier orbitals should be equivalent to my implementation in xtbml, so we can get these quantities from tblite directly and not do the same calculation again in Python, or am I mistaken?

PR #244, opened 2025-04-02
Cleanup the xtbml imports

thfroitzheim:
While adding the new xtbml part to #237, I noticed a few unused or duplicated import statements. Here, I removed them and ordered the import statements. Additionally, I made the function/subroutine definitions consistent with the remaining code.


pitsteinbach (approved):
Thanks for cleaning up, I wasn't aware that we end all subroutine/type code groups again with the name of the subroutine/type. Otherwise I would have taken care of it myself.

Btw, are you using a linter to do the clean-up? If so we could think about making it part of the CI or checkout process.

thfroitzheim:
This I did manually but I agree that a linter would be a great idea.

PR #250, opened 2025-04-11
Replace OMP reductions by thread-private variables and critical summation

thfroitzheim:
Replace reductions on arrays to avoid issues with limited stacksize. Should avoid the need for increased `OMP_STACKSIZE` and `ulimit -s unlimited`. See also [dftd4/dftd4/#287](https://github.com/dftd4/dftd4/pull/287), [grimme-lab/mctc-lib/#79](https://github.com/grimme-lab/mctc-lib/pull/79), [grimme-lab/mctc-lib/#81](https://github.com/grimme-lab/mctc-lib/pull/81), and [grimme-lab/multicharge/#45](https://github.com/grimme-lab/multicharge/pull/45). Moreover, I removed atomic array access in `omp` loops as far as possible. I set the scheduling to static for inherently balanced loops (i.e., identical operations in each iteration),  which allows us to set the runtime to dynamic without any overhead. 

During this, I noticed a small bug in the orbital energies of xtbml due to a lack of initialisation of the output arrays (this leads to a slight change in the tests).


src/tblite/post_processing/xtb-ml/convolution.f90
```diff
@@ -109,7 +109,7 @@ subroutine populate_kernel(self, at, xyz)
    end if
    allocate (self%kernel(nat, nat, n_a), source=0.0_wp)
 
-   !$omp parallel do default(none) collapse(2) &
+   !$omp parallel do schedule(static) default(none) collapse(3) &
```
marvinfriede: Why is this `static`? Setting the value to one is much faster than the calls to `inv_cn`.

src/tblite/post_processing/xtb-ml/density.f90
```diff
@@ -736,33 +749,36 @@ subroutine get_ext_mm_Z(q, dipm, qp, at, xyz, conv, ext_dipm, ext_qp)
...
+            tmp = 1.0_wp / (result * (conv%cn(b, k) + 1.0_wp))
+
+            ext_dipm(:, a, k) = ext_dipm(:, a, k) -r_ab(:) * q(b) * tmp
+
             !sorting of qp xx,xy,yy,xz,yz,zz
-            !$omp atomic
             ext_qp(1, a, k) = ext_qp(1, a, k) + &
-               (1.5_wp * (r_ab(1) * r_ab(1) * q(b))) / (result * (conv%cn(b, k) + one))
-            !$omp atomic
+               1.5_wp * tmp * (r_ab(1) * r_ab(1) * q(b))
```
marvinfriede:
```suggestion
               1.5_wp * tmp * (r_ab(1) * r_ab(1) * q(b))
```

src/tblite/post_processing/xtb-ml/density.f90
```diff
@@ -736,33 +749,36 @@ subroutine get_ext_mm_Z(q, dipm, qp, at, xyz, conv, ext_dipm, ext_qp)
    ext_qp = 0.0_wp
    qp_part = 0.0_wp
 
-   !$omp parallel do default(none) collapse(2)&
-   !$omp private(a, b, k, result, r_ab) shared(ext_dipm, ext_qp, nat, n_a, conv, xyz, q)
+   !$omp parallel do schedule(static) default(none) collapse(2)&
```
marvinfriede:
```suggestion
   !$omp parallel do schedule(static) default(none) collapse(2) &
```

src/tblite/post_processing/xtb-ml/geometry.f90
```diff
@@ -128,17 +128,17 @@ subroutine get_ext_cn(cn, xyz, ext_cn, conv)
    nat = size(cn)
 
    ext_cn = 0.0_wp
-   !$omp parallel do default(none) collapse(2)&
+   !$omp parallel do schedule(static) default(none) collapse(2)&
```
marvinfriede:
```suggestion
   !$omp parallel do schedule(static) default(none) collapse(2) &
```

src/tblite/scf/iterator.f90
```diff
@@ -146,15 +146,29 @@ subroutine get_electronic_energy(h0, density, energies)
...
-   !$omp reduction(+:energies) shared(h0, density) private(spin, iao, jao)
+   ! Thread-private arrays for reduction
+   ! Set to 0 explicitly as the shared variants are potentially non-zero (inout)
+   real(wp), allocatable :: energies_local(:)
+
+   !$omp parallel default(none) &
+   !$omp shared(h0, density, energies) &
+   !$omp private(spin, iao, jao, energies_local)
+   allocate(energies_local(size(energies, 1)), source=0.0_wp)
+   !$omp do schedule(runtime) collapse(3)
```
marvinfriede:
```suggestion
   !$omp do schedule(static) collapse(3)
```

src/tblite/scf/iterator.f90
```diff
@@ -176,16 +190,29 @@ subroutine get_qat_from_qsh(bas, qsh, qat)
...
+   real(wp), allocatable :: qat_local(:, :)
 
    qat(:, :) = 0.0_wp
-   !$omp parallel do schedule(runtime) collapse(2) default(none) &
-   !$omp reduction(+:qat) shared(bas, qsh) private(ish)
+   !$omp parallel default(none) &
+   !$omp shared(bas, qat, qsh) &
+   !$omp private(iat, ish, ispin, qat_local)
+   allocate(qat_local, source=qat)
+   !$omp do schedule(runtime) collapse(2)
```
marvinfriede:
```suggestion
   !$omp do schedule(static) collapse(2) 
```

src/tblite/scf/potential.f90
```diff
@@ -153,7 +153,7 @@ subroutine add_vat_to_vsh(bas, vat, vsh)
    integer :: iat, ish, ii, spin
 
    !$omp parallel do schedule(runtime) collapse(2) default(none) &
-   !$omp reduction(+:vsh) shared(bas, vat) private(spin, ii, ish, iat)
+   !$omp shared(vsh, bas, vat) private(spin, ii, ish, iat)
```
marvinfriede:
```suggestion
   !$omp shared(bas, vat, vsh) private(spin, ii, ish, iat)
```

marvinfriede: Also, schedule `static`?

src/tblite/scf/potential.f90
```diff
@@ -176,7 +176,7 @@ subroutine add_vsh_to_vao(bas, vsh, vao)
    integer :: ish, iao, ii, spin
 
    !$omp parallel do schedule(runtime) collapse(2) default(none) &
-   !$omp reduction(+:vao) shared(bas, vsh) private(ii, iao, ish)
+   !$omp shared(vao, bas, vsh) private(ii, iao, ish)
```
marvinfriede:
```suggestion
   !$omp shared(bas, vao, vsh) private(ii, iao, ish)
```

marvinfriede: Also, schedule `static`?

awvwgk (changes requested):
If we have an atomic access version we don't need to replace it with a local copy, this just increases the memory requirements of the parallel environment. Unless there is a speed benefit, I see no reason for adopting this change here.

src/tblite/coulomb/charge/effective.f90
```diff
@@ -586,10 +605,11 @@ subroutine get_damat_0d(mol, nshell, offset, hubbard, gexp, qvec, dadr, dadL, at
...
       end do
    end do
+   !$omp end do
    !$omp critical (get_damat_0d_)
-   atrace(:, :) = atrace + itrace
-   dadr(:, :, :) = dadr + didr
-   dadL(:, :, :) = dadL + didL
+   atrace(:, :) = atrace(:, :) + itrace(:, :)
+   dadr(:, :, :) = dadr(:, :, :) + didr(:, :, :)
+   dadL(:, :, :) = dadL(:, :, :) + didL(:, :, :)
```
awvwgk, line 612: Please don't create slices on the left hand side of the equation, they should be used on the right hand side to avoid automatic reallocation, but there is no reason to have them on the lhs.

src/tblite/coulomb/charge/gamma.f90
```diff
@@ -545,9 +568,9 @@ subroutine get_damat_3d(mol, nshell, offset, hubbard, rcut, wsc, alpha, qvec, &
    !$omp shared(mol, wsc, alpha, vol, dtrans, rtrans, qvec, hubbard, nshell, offset, rcut) &
    !$omp private(iat, izp, jat, jzp, img, ii, jj, ish, jsh, ui, uj, wsw, vec, dG, dS, &
    !$omp& dGr, dSr, dGd, dSd, itrace, didr, didL)
-   itrace = atrace
-   didr = dadr
-   didL = dadL
+   allocate(itrace, source=atrace)
+   allocate(didr, source=dadr)
+   allocate(didL, source=dadL)
```
awvwgk, line 573: Is there any issue with having the automatic left hand side allocation here?

src/tblite/post_processing/xtb-ml/orbital_energy.f90
```diff
@@ -299,6 +307,9 @@ subroutine get_ehoao_ext(chempot_ext, e_gap_ext, ehoao_ext)
    !> extended HOAO
    real(wp), intent(out) :: ehoao_ext(:, :)
    integer :: a, k
+
```
pitsteinbach, line 310: I haven't profiled those equations yet, but I don't know if they profit from parallelization, since the k loop is over spin which is for most cases still 1. So we only iterate over number of atoms, so the OMP overhead might not be worth it.

src/tblite/wavefunction/mulliken.f90
```diff
@@ -158,7 +197,7 @@ end subroutine get_molecular_quadrupole_moment
 
 
 !> Evaluate Wiberg/Mayer bond orders
-subroutine get_mayer_bond_orders(bas, smat, pmat, mbo)
+subroutine get_mayer_bond_orders(bas, smat, pmat, mbo, unpaired)
```
pitsteinbach, line 200: that's a good solution to avoid code duplication

src/tblite/coulomb/charge/effective.f90
```diff
@@ -547,6 +564,8 @@ subroutine get_damat_0d(mol, nshell, offset, hubbard, gexp, qvec, dadr, dadL, at
 
    integer :: iat, jat, izp, jzp, ii, jj, ish, jsh
    real(wp) :: vec(3), r1, gam, arg, dtmp, dG(3), dS(3, 3)
+
+   ! Thread-private array for reduction
    real(wp), allocatable :: itrace(:, :), didr(:, :, :), didL(:, :, :)
```
pitsteinbach, line 569: do we want to keep the naming scheme and call these also `_local` or not in this case since there is no global value for the gradient intermediate?

src/tblite/post_processing/xtb-ml/density.f90
```diff
@@ -625,44 +644,38 @@ subroutine get_ext_mm(q, dipm, qp, at, xyz, conv, ext_dipm, ext_qp)
...
-               (result * (conv%cn(b, k) + one))
-            !$omp atomic
-            ext_qp(5, a, k) = ext_qp(5, a, k) + &
-               (1.5_wp * (-one * (r_ab(3) * dipm(2, b) + r_ab(2) * dipm(3, b)) + r_ab(2) * r_ab(3) * q(b))) / &
-               (result * (conv%cn(b, k) + one))
-            !$omp atomic
-            ext_qp(6, a, k) = ext_qp(6, a, k) + &
-               (1.5_wp * (-one * (r_ab(3) * dipm(3, b) + r_ab(3) * dipm(3, b)) + r_ab(3) * r_ab(3) * q(b))) / &
-               (result * (conv%cn(b, k) + one))
+            ext_qp(1, a, k) = ext_qp(1, a, k) + 1.5_wp * tmp * &
```
pitsteinbach, line 660: isn't there potential for data races since multiple `b`-valued threads can access the same `a`-valued shared value? Same for the other instances where atomic was removed

pitsteinbach: We could do the same "trick" then in other subroutines and have a private `qp_intermediate` which is added to the shared array after completing the inner loop?

test/unit/test_xtbml.f90
```diff
@@ -723,10 +723,10 @@ subroutine test_orbital_energy_ref(error)
       "ext_chem_pot_alpha = [ -0.9205182203755125, -0.9205182203755127, -0.9334269287974830 ]", &
       "ext_HOAO_alpha = [ -0.9232658053262217, -0.9232658053262219, -0.9368077181575476 ]", &
       "ext_LUAO_alpha = [ -0.9177706354248032, -0.9177706354248034, -0.9300461394374184 ]", &
-      "ext_gap_beta = [ 9.4593893126801003, 9.4593893126801003, 6.5604665198186094 ]", &
-      "ext_chem_pot_beta = [ -12.9537156502782160, -12.9537156502782196, -13.9314158798429979 ]", &
-      "ext_HOAO_beta = [ -17.6834103066182671, -17.6834103066182706, -17.2116491397523035 ]", &
-      "ext_LUAO_beta = [ -8.2240209939381650, -8.2240209939381685, -10.6511826199336923 ]", &
+      "ext_gap_beta = [ 9.4538889011103251, 9.4538889011103251, 6.5537012263493972 ]", &
```
pitsteinbach, line 726: thanks for fixing these!

pitsteinbach (commented):
In addition to my comments, as we already exchanged, critical comes with an overhead compared to atomic access. Whereas the latter can also be hardware-accelerated.

src/tblite/coulomb/charge/effective.f90
```diff
@@ -258,9 +258,16 @@ subroutine get_amat_0d(mol, nshell, offset, hubbard, gexp, amat)
...
+
+   !$omp parallel default(none) &
    !$omp shared(amat, mol, nshell, offset, hubbard, gexp) &
-   !$omp private(iat, izp, ii, ish, jat, jzp, jj, jsh, gam, vec, r1, r1g, tmp)
+   !$omp private(iat, izp, ii, ish, jat, jzp, jj, jsh, gam, &
+   !$omp& vec, r1, r1g, tmp, amat_local)
+   allocate(amat_local(size(amat, 1), size(amat, 2)), source=0.0_wp)
+   !$omp do schedule(runtime)
    do iat = 1, mol%nat
       izp = mol%id(iat)
```
foxtran, line 272: Such changes affect performance of build with disabled OpenMP. One needs to avoid extra allocations for noOMP builds.

src/tblite/coulomb/charge/effective.f90
```diff
@@ -258,9 +258,16 @@ subroutine get_amat_0d(mol, nshell, offset, hubbard, gexp, amat)
...
+   ! Set to 0 explicitly as the shared variants are potentially non-zero (inout)
+   real(wp), allocatable :: amat_local(:, :)
+
+   !$omp parallel default(none) &
    !$omp shared(amat, mol, nshell, offset, hubbard, gexp) &
-   !$omp private(iat, izp, ii, ish, jat, jzp, jj, jsh, gam, vec, r1, r1g, tmp)
+   !$omp private(iat, izp, ii, ish, jat, jzp, jj, jsh, gam, &
+   !$omp& vec, r1, r1g, tmp, amat_local)
+   allocate(amat_local(size(amat, 1), size(amat, 2)), source=0.0_wp)
+   !$omp do schedule(runtime)
```
foxtran, line 270: Is `runtime` a good scheduler?

awvwgk:
Closing this, we can revisit this in the future if needed.

PR #252, opened 2025-04-20
Add support for reading and writing npy and npz files

awvwgk:
Implement support for reading and writing `npy` and `npz` format.

- `npy` format is defined by a header describing the binary data
  - Implementation of `npy` format based on https://github.com/fortran-lang/stdlib/pull/581
- `npz` format is defined as zip file of `npy` files
  - local header, global header and footer formatting based on https://en.wikipedia.org/wiki/ZIP_(file_format)
  - CRC32 checksum based on https://fortranwiki.org/fortran/show/hash+value+example
  - Only supports uncompressed inputs (compression flag == 0)
- allow saving wavefunctions in `npz` format
  - use `tblite0_` prefix as versioning in case of future incompatible changes
- allow loading wavefunctions from `npz` format
  - rigorously validate consistency of dimensions
  - allow passing in optional dimension attributes to check for mismatches
- add `--restart` / `--no-restart` (default) option
  - keep `--no-restart` as default for now, change to `--restart` as default later


awvwgk:
Pit @pitsteinbach, Thomas @thfroitzheim, please have a look and let me know if you have any comments and suggestions.

app/cli_help.f90
```diff
@@ -116,6 +116,10 @@ module tblite_cli_help
       "      --etemp <real>       Electronic temperature for calculation (Default: 300K)"//nl//&
       "      --guess <name>       Guess for the initial populations, possible options:"//nl//&
       "                           sad (default), eeq, ceh (Charge-Extended Hückel method)"//nl//&
+      "      --restart [<file>]   Restart calculation from previous run"//nl//&
```
thfroitzheim, line 119: Maybe we should add a notice that the `--restart` flag overwrites the `--guess` or add a warning if both are used at the same time to avoid confusion

src/tblite/wavefunction/restart.f90
```diff
@@ -0,0 +1,226 @@
...
+
+   if (stat == 0) then
+      call load_npz(filename, prefix//"dpat", wfn%dpat, stat, msg)
+   end if
+   if (stat == 0) then
+      call load_npz(filename, prefix//"qpat", wfn%qpat, stat, msg)
+   end if
+
+   if (stat /= 0) then
+      if (.not.allocated(msg)) msg = "Failed to read wavefunction from '"//filename//"'"
```
thfroitzheim, line 99: Maybe we should add the version prefix to the error message as it is still quite readable in the final file.

test/cli/01a-gfn1xtb.resp
```diff
@@ -1,3 +1,5 @@
 $ORIGIN/01-ala-xab.xyz
+--restart
+01-ala-xab-restart.npz
```
thfroitzheim, line 3: This test currently only writes the restart file. We could also include one with the restart file.

thfroitzheim (commented):
Only a few small points. From what I tested it works great.

src/tblite/wavefunction/restart.f90
awvwgk: I would consider the version prefix an implementation detail on our side. I don't see necessity to expose the user to it. Once we make a breaking change and have the option to read `tblite0_` prefixed data as a recovery option, it would make sense to highlight older versions in the error message.

app/cli_help.f90
awvwgk: Successfully reading a restart file should disable any guess as a default behavior. Requesting a restart will still fall back to the guess in case we encounter an unusable restart file or the restart file does not exist.

test/cli/01a-gfn1xtb.resp
awvwgk: Yes, this mainly tests writing. We need a bit more infrastructure for testing to handle restart files. I would keep this for the a future iteration.

PR #254, opened 2025-04-23
Fix CRC32 compilation with xtb: set proper sizes of integers instead default one

foxtran:
Before the patch, compiling of xtb with ifx/icx was not possible since compilation of current master branch of tblite generates errors:
```
xtb-latest/build_noomp/_deps/tblite-src/src/tblite/io/numpy/crc32.f90(38): error #6366: The shapes of the array expressions do not conform.   [CRC_TABLE]
   integer(i4), parameter :: crc_table(0:255) = transfer([ &
-----------------------------^
```

The problem comes from using -fdefault-integer-8 / -fdefault-real-8 (for GNU, but anyway) flags passed during xtb compilation, and therefore sizes of arrays was incorrent with these flags. At the same time, building tblite does not gives this error since it does not use these flags.

### DIFF
--- a/src/tblite/io/numpy/crc32.f90
+++ b/src/tblite/io/numpy/crc32.f90
@@ -34,8 +34,7 @@ module tblite_io_numpy_crc32
       module procedure crc32_hash_rdp_r1
    end interface crc32_hash
 
-   integer(i4), parameter :: crc_stub(256) = 0
-   integer(i4), parameter :: crc_table(0:255) = transfer([ &
+   integer(i4), parameter :: crc_table(0:255) = (/ &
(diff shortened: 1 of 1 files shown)


awvwgk (approved):
Looks good to me. I used `int(z'...', i4)` before, but run into an issue with GFortran requiring `-fno-range-check` or similar because of signed integer overflows. Using `real(z'...', sp)` with casting into `i4` worked around this. If it works here in CI, we can go ahead and use it.

foxtran:
> but run into an issue with GFortran requiring -fno-range-check or similar because of signed integer overflows

That is strange since BOZ constants does not have signs :| Looks like a bug in compiler.

foxtran:
@awvwgk, should I rebase?

PR #255, opened 2025-04-23
Add Intel-LLVM support for meson

foxtran:
Added support of Intel-LLVM for meson (CMake and FPM works fine) as well as testing of this combination in CI.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -71,8 +71,7 @@ jobs:
           build-type: debug
           compiler: intel
           version: none
-          toolchain:
-          - {compiler: intel-classic, version: '2021.10'}
+          toolchain: {compiler: intel-classic, version: '2021.10'}
 
         - os: ubuntu-latest
           build: fpm
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,21 +66,21 @@ jobs:
           build-type: debug
           compiler: intel
           version: none
-          toolchain: {compiler: intel-classic, version: '2021.10'}
+          toolchain: {compiler: intel-classic, version: '2021.10', mkl_version: '2021.2.0'}
 
         - os: ubuntu-latest
           build: meson
           build-type: debug
           compiler: intel
           version: none
-          toolchain: {compiler: intel, version: '2024.1'}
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -289,6 +289,50 @@ jobs:
...
+      APT_PACKAGES: >-
+        intel-oneapi-mkl-${{ matrix.toolchain.version }} intel-oneapi-mkl-devel-${{ matrix.toolchain.version }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Setup Fortran
+        uses: fortran-lang/setup-fortran@v1
+        with:
+          compiler: ${{ matrix.toolchain.compiler }}
+          version: ${{ matrix.toolchain.version }}
```
awvwgk: Would be great if we can merge this with the existing build stage

foxtran: I need to have a look how to do it. Not sure about today.

foxtran: Ok. It works now :D

awvwgk (approved):
Thanks for updating the build pipeline

awvwgk:
Looks like this is failing on main now...
````
The Meson build system
Version: 1.7.2
Source dir: /home/runner/work/tblite/tblite
Build dir: /home/runner/work/tblite/tblite/_build
Build type: native build
Project name: tblite
Project version: 0.4.0

meson.build:17:0: ERROR: Unknown compiler(s): [['ifort']]
The following exception(s) were encountered:
Running `ifort --help` gave "[Errno 2] No such file or directory: 'ifort'"
Running `ifort --version` gave "[Errno 2] No such file or directory: 'ifort'"
Running `ifort -V` gave "[Errno 2] No such file or directory: 'ifort'"

A full log can be found at /home/runner/work/tblite/tblite/_build/meson-logs/meson-log.txt
````

foxtran:
Hm... Looks like a problem with cache.

foxtran:
I do not see the reason why intel compilers were not cached. There is their location:

https://github.com/fortran-lang/setup-fortran/blob/a6ee71630f0467c94aa4c5b677fb618978362cfb/setup-fortran.sh#L371

toxtran:
Intel compilers are in cache. The simplest thing which we can do - disable caching then it will work.

PR #256, opened 2025-04-24
Use npz format to store post-processing results

awvwgk:
- cleanup double dictionary implementation
- use npz format for serializing numerical data

### DIFF
--- a/app/cli.f90
+++ b/app/cli.f90
@@ -72,6 +72,8 @@ module tblite_cli
       type(solvation_input), allocatable :: solvation
       !> Input for post processing container
       character(len=:), allocatable :: post_processing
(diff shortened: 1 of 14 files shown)


pitsteinbach:
Why would we remove the toml output and input functionality? 
How about having both npz and toml, and use the corresponding parser or lexer depending on file name ending?

awvwgk:
TOML is more a configuration format than a serialization format. For input files I strongly prefer TOML as format, for outputs however I think either JSON or other formats optimized for numeric data are preferable, like HDF5 or NPZ. For the post processing outputs, TOML seemed like a stop-gap solution, now we have a better alternative and I prefer to use it going forward.

PR #259, opened 2025-04-25
Fix flicking overlap test

### LINKED ISSUE #233: Flicking test: tblite:unit / integral-overlap
See https://github.com/tblite/tblite/actions/runs/13637501302/job/38119750678?pr=232#step:18:32

How to reproduce CI: https://github.com/grimme-lab/xtb/issues/1182

foxtran:
Fixes #233

The problem comes from a bit strange expression for norming vectors to specific distance and therefore obtained vectors could be too short for enough precise numerical integration that lead to flicking overlap tests. Now, vectors will have the proper length according documented values. I have also expanded test sets.

### DIFF
--- a/test/unit/test_integral_overlap.f90
+++ b/test/unit/test_integral_overlap.f90
@@ -1879,6 +1879,13 @@ subroutine test_overlap_diat_grad_ss_z(error)
 
    call test_overlap_diat_grad_gen(vec, ksig, kpi, kdel, cgtoi, cgtoj, error)
 
+   ! Vector nearly along the z-axis to test the ill-defined gradient
+   vec(1) = 1e-7_wp
+   vec(2) = 1e-7_wp
+   vec(3) = 0.5_wp
+
+   call test_overlap_diat_grad_gen(vec, ksig, kpi, kdel, cgtoi, cgtoj, error)
+
 end subroutine test_overlap_diat_grad_ss_z
 
 subroutine test_overlap_diat_grad_sp(error)
@@ -1949,6 +1956,13 @@ subroutine test_overlap_diat_grad_sp_z(error)
 
    call test_overlap_diat_grad_gen(vec, ksig, kpi, kdel, cgtoi, cgtoj, error)
 
+   ! Vector nearly along the z-axis to test the ill-defined gradient
+   vec(1) = 1e-7_wp
+   vec(2) = 1e-7_wp
+   vec(3) = 0.5_wp
+
+   call test_overlap_diat_grad_gen(vec, ksig, kpi, kdel, cgtoi, cgtoj, error)
+
 end subroutine test_overlap_diat_grad_sp_z
 
 subroutine test_overlap_diat_grad_pp(error)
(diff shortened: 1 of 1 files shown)


test/unit/test_integral_overlap.f90
```diff
@@ -2125,15 +2205,29 @@ subroutine test_overlap_diat_grad_dd_z(error)
    call slater_to_gauss(ng, 4, 2, 1.0_wp, cgtoi, .true., stat)
    call slater_to_gauss(ng, 3, 2, 1.3_wp, cgtoj, .true., stat)
 
-   ! Vector along the z-axis to test the ill-defined gradient
-   vec(1) = -4.2738582800000008E-005_wp
-   vec(2) = 1.9619373507000001E-004_wp
-   vec(3) = -5.1529538776373203_wp
-
```
foxtran, line 2132: There was a test with distorted coordinates which are close to z-axis. I deleted it.

awvwgk (approved):

test/unit/test_integral_overlap.f90
thfroitzheim: There was actually an idea behind this test. The orientation almost along the z axis lead to a flip from one side of the z-axis to the other, which lead in an earlier version to an error. But currently these value don't test this case any more, since I seem to have reduced the stepsize in the numerical gradient to `1e-6`. 

Still, this is a relevant case where the diatomic frame transformation can fail. Let's instead add one test with a vector like this: 
```
vec(1) = 1E-7_wp
vec(2) = 1E-7_wp
vec(3) = 0.5_wp
```

test/unit/test_integral_overlap.f90
```diff
@@ -1849,15 +1854,29 @@ subroutine test_overlap_diat_grad_ss_z(error)
...
-   ! Vector along the z-axis to test the ill-defined gradient
-   vec(1) = 0.0_wp
-   vec(2) = 0.0_wp   
-   vec(3) = 0.5_wp
-
    ksig = 0.1_wp
    kpi = 0.2_wp
    kdel = 0.5_wp
 
+   ! Vector along the x-axis to test the ill-defined gradient
```
thfroitzheim: The `x` and `y` direction actually cannot be ill-defined. The reason is the special status of the z-axis in the spherical coordinate system (i.e. if theta is zero or 180 degrees the x and y components become ill-defined but if theta is 90 degrees the z component and either x or y components remain defined). 

So I would still only focus on the z component (this is the case for all components ss, sp, pp, ...). 

It would make more sense to always add one test which flicks from one side of the z axis to the other

thfroitzheim (changes requested):
Good catch that I missed the `sqrt` in the normalization. We can actually also properly transform very short vectors but then we need a smaller stepsize

test/unit/test_integral_overlap.f90
foxtran: > The x and y direction actually cannot be ill-defined.

I left those test if one decide to change axis order :-)

> It would make more sense to always add one test which flicks from one side of the z axis to the other

Added.

thfroitzheim (approved):

PR #262, opened 2025-05-02
Move density build to solver object

pitsteinbach:
In order to prepare for solvers which are not based on diagonalization, I propose as done before by @awvwgk to move all the functions relevant to building the density matrix to the solver type object, this allows a later overwrite by a solver object that uses DMP methods for example.

Another functionality introduced by this PR is that the solver might not be deleted at the end of an SCF cycle if a flag within the context is changed. The reason for that is that for applications such as CREST or geometry optimizations in general, we would want to reuse the solver instance as to not pay the setup time too often. This is especially helpful in the context of  GPU based methods as it allows to reuse buffers allocated on the GPU in latter SCF cycles.

Relating to #71.

### DIFF
--- a/src/tblite/context/type.F90
+++ b/src/tblite/context/type.F90
@@ -46,11 +46,9 @@ module tblite_context_type
       !> Optional logger to be used for writing messages
       class(context_logger), allocatable :: io
       !> Optional factory for creating electronic solvers
-      class(context_solver), allocatable :: ctxsolver
+      class(context_solver), allocatable :: solver
       !> Color support for output
       type(context_terminal) :: terminal = context_terminal()
-      !> Solver instance moved here to keep MD overhead low
-      class(solver_type), allocatable :: solver
    contains
       !> Write a message to the output
       procedure :: message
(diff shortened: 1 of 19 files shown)


src/tblite/context/type.F90
```diff
@@ -44,9 +46,11 @@ module tblite_context_type
       !> Optional logger to be used for writing messages
       class(context_logger), allocatable :: io
       !> Optional factory for creating electronic solvers
-      class(context_solver), allocatable :: solver
+      class(context_solver), allocatable :: ctxsolver
```
awvwgk: Can we keep the original name? The ctx prefix hete does not seem necessary

pitsteinbach: then I would need to rename the solver 4 lines down from this line. I wanted to have the actual solver as part of the context since this object usually persists in implementations.

awvwgk: I think the `context_solver` is already flexible enough to allow keeping persistent memory or implementation details between runs. I would like to keep the current logic of instantiating a new solver in every singlepoint from the `context_solver`. You can use a pointer in the created solver which references the persistent context memory for example.

pitsteinbach: Okay, I see your point,  I will transfer the setup logic to the context solver.
Can you elaborate on why you think having this overhead in creating a solver every time makes sense?

src/tblite/lapack/solver.f90
```diff
@@ -57,13 +58,15 @@ module tblite_lapack_solver
 
 
 !> Create new electronic solver
-subroutine new(self, solver, ndim)
+subroutine new(self, solver, ndim, acc)
```
foxtran, line 61: Does it make sense to have an extra acc argument instead of set_accuracy/get_accuracy routines? It could be useful to redefine acc during geometry optimizations.

pitsteinbach: Since we create a new solver instance per SCF, an application can rely on changing the accuracy per singlepoint, through the singlepoint function call. An alternative would be to make the accuracy part of the context, but I think we should try and keep the API to singlepoint stable.

pitsteinbach:
in order  to run the whole CI I added some fixes regarding the newest meson version giving errors.
@awvwgk do you want me to open a separate PR for that, how to we want to address this situation? 
Other suggestions for this PR?

marvinfriede:
> in order to run the whole CI I added some fixes regarding the newest meson version giving errors. @awvwgk do you want me to open a separate PR for that, how to we want to address this situation? Other suggestions for this PR?

I would prefer a separate PR. Otherwise, we have to wait for this PR to be merged.
Also, can you only exclude the broken version without excluding later version (`meson!=1.8.0`)?

pitsteinbach:
@awvwgk can we move forward with this or do you want more changes?

PR #263, opened 2025-05-05
Fix CI by excluding version 1.8.0

pitsteinbach:
Meson Release 1.8.0 leads to the CI failing currently, this is should be a temporary patch until meson has been bumped to a version that allows stable compiling again.

To address this I exclude the meson version in use for the CI containers as well as setting meson explicitly as a requirement in `pyproject.toml` instead of relying on `meson-python`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 env:
   BUILD_DIR: _build
   PIP_PACKAGES: >-
-    meson
+    meson<1.8.0 
     cmake
     ninja
     gcovr
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 env:
   BUILD_DIR: _build
   PIP_PACKAGES: >-
-    meson<1.8.0 
+    "meson!=1.8.0" 
     cmake
     ninja
     gcovr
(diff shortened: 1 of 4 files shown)


assets/ci/python-env.yaml
```diff
@@ -14,4 +14,4 @@ dependencies:
   - ase
   - qcelemental
   - matplotlib-base
-  - meson
+  - 'meson!=1.8.0'
```
marvinfriede:
```suggestion
  - meson!=1.8.0
```

assets/ci/build-env.yaml
```diff
@@ -2,7 +2,7 @@ name: devel
 channels:
   - conda-forge
 dependencies:
-  - meson
+  - 'meson!=1.8.0'
```
marvinfriede, line 5:
```suggestion
  - meson!=1.8.0
```

.github/workflows/build.yml
```diff
@@ -5,7 +5,7 @@ on: [push, pull_request]
 env:
   BUILD_DIR: _build
   PIP_PACKAGES: >-
-    meson
+    "meson!=1.8.0"
```
marvinfriede, line 8:
```suggestion
    meson!=1.8.0
```

assets/ci/build-env.yaml
pitsteinbach: for conda I needed the ' I think.

marvinfriede: I think only if you directly install something with version restrictions from command line. The YAML file does not need strings. Installation from the file works without strings, I just tested this.

pitsteinbach:
yeah @awvwgk opened basically the same PR I am closing this then.

Issue #261, opened 2025-04-29
Building with GCC 9 fails without `-fno-range-check`


marvinfriede:
### Bug

With #254, compilation with GCC9 fails. Not sure if this warrants a fix, since compilation seems to work with other compilers (GCC10+, ifx).
Otherwise, this should just serve as a reference if someone else encounters this issue.

### Solution

Compile with `-fno-range-check`, e.g., `fpm build --flag="-fno-range-check"`.

### Error Message

Meson:

```sh
[288/512] Compiling Fortran object libtblite.so.0.4.0.p/src_tblite_io_numpy_crc32.f90.o
FAILED: libtblite.so.0.4.0.p/src_tblite_io_numpy_crc32.f90.o libtblite.so.0.4.0.p/tblite_io_numpy_crc32.mod 
gfortran -Ilibtblite.so.0.4.0.p -I. -I.. -Isubprojects/mctc-lib/libmctc-lib.a.p -Isubprojects/json-fortran-8.2.5/libjsonfortran.a.p -Isubprojects/dftd4/libdftd4.a.p -I../subprojects/dftd4/include -Isubprojects/multicharge/libmulticharge.a.p -Isubprojects/s-dftd3/libs-dftd3.a.p -I../subprojects/s-dftd3/include -Isubprojects/toml-f/libtoml-f.a.p -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -O3 -ffree-line-length-none -fbacktrace -Wno-maybe-uninitialized -fPIC -fopenmp -fopenmp -Jlibtblite.so.0.4.0.p -o libtblite.so.0.4.0.p/src_tblite_io_numpy_crc32.f90.o -c ../src/tblite/io/numpy/crc32.f90
../src/tblite/io/numpy/crc32.f90:101:34:

  101 |       & int(z'b40bbe37', i4), int(z'c30c8ea1', i4), int(z'5a05df1b', i4), int(z'2d02ef8d', i4)]
      |                                  1
Error: Arithmetic overflow converting INTEGER(16) to INTEGER(4) at (1). This check can be disabled with the option ‘-fno-range-check’
../src/tblite/io/numpy/crc32.f90:197:33:

  197 |       crc = ieor(shiftr(crc, 8), crc_table(iand(ieor(crc, iachar(chunk(1))), 255)))
      |                                 1
Error: Function ‘crc_table’ at (1) has no IMPLICIT type
[...]

PR #264, opened 2025-05-05
Deprecation warning for GCC 9, and adding compiler flag

pitsteinbach:
As addressed by #261, building under GCC 9 and its Fortran compiler fails without adding the `-fno-range-check` flag. Since GCC 9 is the default compiler for Ubuntu LTS 20.04 which is still commonly used I would add a warning and add the flag to the compilation. 

Here is my suggestion, for those also on 20.04 please check with your cmake and meson versions.

### DIFF
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -38,6 +38,17 @@ set(
 )
 set(module-dir "${module-dir}" PARENT_SCOPE)
 
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
+  if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 10.0)
+    message(
+      WARNING
+      "GNU Fortran version < 10.0 is not fully supported. Please use a newer version if you can."
+    )
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS} -fno-range-check")
+    set(CMAKE_Fortran_FLAGS "${CMAKE_Fortran_FLAGS}" PARENT_SCOPE)
+  endif()
+endif()
+
 # Set build type as CMake does not provide defaults
 if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
   set(
--- a/config/CMakeLists.txt
+++ b/config/CMakeLists.txt
@@ -42,7 +42,7 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
   if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 10.0)
     message(
       WARNING
-      "GNU Fortran version < 10.0 is not fully supported. Please use a newer version if you can."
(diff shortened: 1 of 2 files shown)


config/meson.build
```diff
@@ -29,6 +29,14 @@ if fc_id == 'gcc'
     '-Wno-maybe-uninitialized',
     language: 'fortran',
   )
+  message('GCC version: ' + fc.version())
```
awvwgk:
```suggestion
```

config/meson.build
```diff
@@ -29,6 +29,14 @@ if fc_id == 'gcc'
...
+  if fc.version().contains('9.')
```
awvwgk: https://mesonbuild.com/Reference-manual_elementary_str.html#strversion_compare
```suggestion
  if fc.version().version_compare('<10')
```

awvwgk (changes requested):

config/meson.build
```diff
@@ -29,6 +29,14 @@ if fc_id == 'gcc'
...
+    message('GNU Fortran version < 10.0 is not fully supported. Please use a newer version if you can.')
```
marvinfriede: I think the message is a bit confusing. The flag will be set automatically (and silently) and tblite will compile and work without further issues. The user won't know what "not fully supported" means and if there is an impact on the functionality.

pitsteinbach:
```suggestion
    warning('tblite is not tested for GNU Fortran version < 10.0, functionality might be impacted. Please use a newer version if you can.') 
```

pitsteinbach: Similar to cmake we can use a warning and this type of message. Since we don't test we don't know if there is impact on the functionality.

marvinfriede: Let's inform the user about the additional flag. What about:

```sh
warning("tblite it not tested with GNU Fortran < 10, functionality might be impacted. The build automatically adds ‘-fno-range-check’ for compatibility. Please upgrade to a newer compiler version if possible.")
```

pitsteinbach:
```suggestion
    message('tblite it not tested with GNU Fortran < 10, functionality might be impacted. The build automatically adds ‘-fno-range-check’ for compatibility. Please upgrade to a newer compiler version if possible.') 
```

config/CMakeLists.txt
```diff
@@ -38,6 +38,17 @@ set(
 )
 set(module-dir "${module-dir}" PARENT_SCOPE)
 
+if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
+  if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 10.0)
+    message(
+      WARNING
+      "GNU Fortran version < 10.0 is not fully supported. Please use a newer version if you can."
```
pitsteinbach:
```suggestion
      "tblite it not tested with GNU Fortran < 10, functionality might be impacted. The build automatically adds ‘-fno-range-check’ for compatibility. Please upgrade to a newer compiler version if possible."
```

PR #265, opened 2025-05-05
Exclude meson version 1.8.0

awvwgk:
See https://github.com/mesonbuild/meson/issues/14528
See https://github.com/mesonbuild/meson/issues/14532
See https://github.com/mesonbuild/meson/pull/14549

### DIFF
--- a/python/meson.build
+++ b/python/meson.build
@@ -19,7 +19,7 @@ project(
   'c',
   version: '0.4.0',
   license: 'LGPL-3.0-or-later',
-  meson_version: '>=0.57.2',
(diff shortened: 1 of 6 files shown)


pitsteinbach:
see #263

python/meson.build
```diff
@@ -19,7 +19,7 @@ project(
   'c',
   version: '0.4.0',
   license: 'LGPL-3.0-or-later',
-  meson_version: '>=0.57.2',
```
foxtran, line 22: Looks like comma-separated version constraint is not allowed before 1.8.0 :D

Issue #287, opened 2025-07-31
Using --post-processing xtbml with CPCM solvation causes segmentation fault


ioannoid:
```
xTB calculator
 | screened Coulomb repulsion
 | Coulomb electrostatics
 |  * isotropic Klopman-Ohno-Mataga-Nishimoto electrostatics
 |  * anisotropic electrostatics
 |  * onsite third-order electrostatics
 | self-consistent DFT-D4 dispersion
 | Interactions (1)
 | [1] polarizable continuum model

repulsion energy          1.8365738085632E+00 Eh
dispersion energy         6.4067400169765E-03 Eh
interaction energy        0.0000000000000E+00 Eh
number of electrons       2.4600000000000E+02 e
integral cutoff           4.0000000000000E+01 bohr

------------------------------------------------------------
  cycle        total energy    energy error   density error
------------------------------------------------------------
      1     -128.3167360174  -1.3015972E+02   1.1549908E-01
      2     -128.7360240000  -4.1928798E-01   7.0097422E-02
      3     -128.9512479511  -2.1522395E-01   3.6874570E-02
      4     -128.8829990813   6.8248870E-02   3.6914989E-02
      5     -129.0079519948  -1.2495291E-01   1.1363336E-02
      6     -129.0033331534   4.6188414E-03   5.4628569E-03
      7     -128.9983401284   4.9930250E-03   1.7452850E-03
      8     -128.9980574932   2.8263524E-04   8.5013214E-04
      9     -128.9987463152  -6.8882206E-04   4.4735233E-04
     10     -128.9989431082  -1.9679299E-04   1.0261677E-04
     11     -128.9989704925  -2.7384237E-05   4.0328439E-05
     12     -128.9989699410   5.5142002E-07   2.2039268E-05
     13     -128.9989649129   5.0281658E-06   9.7145386E-06
     14     -128.9989606284   4.2844337E-06   4.4305753E-06
     15     -128.9989595715   1.0569681E-06   2.8326787E-06
     16     -128.9989597045  -1.3303034E-07   3.0779209E-07
------------------------------------------------------------

electronic energy        -1.3084194025308E+02 Eh
total energy             -1.2899895970450E+02 Eh


Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x155554c54daf in ???
#1  0x155555406940 in __tblite_solvation_cpcm_MOD_get_energy
        at ../src/tblite/solvation/cpcm.f90:218
[...]

PR #290, opened 2025-08-08
Fix CPCM xtbml energy bug

pitsteinbach:
As described by #287, we get a segmentation fault when running the `xtbml` post-processing features  together with CPCM. This comes from the cache update function of CPCM with a subsequent call to `get_energy` without first calling `get_potential`. I fixed it by changing the order in which I call these functions.

However, it might be an idea to take a look at the update function of CPCM, since I don't think that we need to deallocate the cache every time we call update. @awvwgk you implemented this solvation, can you take a look, since ALPB for example doesn't use this deallocate everything strategy.


awvwgk (approved):
Thanks for sharing. Looks good to me

PR #291, opened 2025-09-01
Compressed restart from Mulliken shell charges and atomic multipoles (C/Python API, spin support, tests)

MauriceDHanisch:
## Summary
Adds a lightweight restart mechanism that seeds SCF from Mulliken shell charges together with atomic dipole and quadrupole moments. Provides C and Python APIs (including spin-resolved support), wires the guess into wavefunction initialization, and adds tests and a Python example. Default behavior is unchanged unless a guess is explicitly provided.

## Motivation
- Reduce restart payload size vs. full density/coefficients.
- Enable portable restarts across language bindings (C, Python).
- Support both closed-shell and open-shell calculations.

## Key changes
- Wavefunction seeding from compact restart data:
  - If a guess is provided, initialize `wfn%qsh`, `wfn%dpat`, `wfn%qpat`, and derive `wfn%qat` via `get_qat_from_qsh`.
  - Strict shape checks; all three arrays are required together.
  - Implementation in `src/tblite/api/calculator.f90` (`check_wavefunction` and call site).

- Result container extended to hold restart guess:
  - `charges_guess_shell [nsh, nspin]`, `dpat_guess [3, nat, nspin]`, `qmat_guess [6, nat, nspin]`.
  - Allocation/population added in C-bindings.
  - Implementation in `src/tblite/api/result.f90`.

- C API:
  - `tblite_set_result_shell_charges_and_moments_guess(...)` (closed-shell).
  - `tblite_set_result_shell_charges_and_moments_guess_spin(...)` (spin-resolved).
  - Declarations in `include/tblite/result.h`, implementations in `src/tblite/api/result.f90`.

- Python API:
  - `Result.set("shell-charges-and-moments-guess", (qsh, dpat, qmat))`.
  - Shape normalization and dispatch to the appropriate C API (closed-shell vs. spin).
  - Implemented in `python/tblite/interface.py` and `python/tblite/library.py`.

- Tests and example:
  - New C test `test_calc_restart_compressed` in `test/api/main.c`: compares energy from compressed restart vs. full run (closed-shell).
  - Example script `restart_from_charges.py`: demonstrates compressed restart and compares to a full wavefunction restart; includes a spin-polarized path.

## API additions
- C:
  - `tblite_set_result_shell_charges_and_moments_guess(error, res, qsh, nsh, dpat, nat, qmat, nat2)`
  - `tblite_set_result_shell_charges_and_moments_guess_spin(error, res, qsh, nsh, dpat, nat, qmat, nat2, nspin)`
- Python:
  - `Result.set("shell-charges-and-moments-guess", (qsh, dpat, qmat))`
  - Accepted shapes:
    - `qsh`: `(nsh,)` or `(2, nsh)`
    - `dpat`: `(nat, 3)`/`(3, nat)` or `(2, nat, 3)`/`(2, 3, nat)`
    - `qmat`: `(nat, 6)`/`(6, nat)` or `(2, nat, 6)`/`(2, 6, nat)`

## Behavior and validation
- Providing `qsh` also requires providing `dpat` and `qmat`; otherwise, a fatal error is raised via the context.
- Shapes are validated against the system (`nat`, `nsh`) and active spin count.
- Without a guess, existing initialization is unchanged.

## Backward compatibility
- No breaking changes. Behavior is identical unless the new setter APIs are used.

## Affected files
- `src/tblite/api/calculator.f90` (consume guess, seed `wfn`)
- `src/tblite/api/result.f90` (store and set guess; C bindings)
- `include/tblite/result.h` (C API declarations)
- `python/tblite/interface.py` (setter wiring)
- `python/tblite/library.py` (shape normalization, C calls)
- `test/api/main.c` (new compressed-restart test)

## Tests
- `test_calc_restart_compressed` verifies energy agreement (closed-shell) when restarting from `(qsh, dpat, qmat)`.

## Notes
- CLI behavior unchanged.
- Documentation and changelog entries can be added in a follow-up.


src/tblite/api/calculator.f90
```diff
@@ -560,7 +561,17 @@ subroutine get_singlepoint_api(vctx, vmol, vcalc, vres) &
 
    call check_results(res%results)
    
-   call check_wavefunction(res%wfn, mol%ptr, calc%ptr, calc%etemp, calc%nspin, calc%guess)
+   if (allocated(res%charges_guess_shell)) then
+      if (.not.allocated(res%dpat_guess) .or. .not.allocated(res%qmat_guess)) then
+         call fatal_error(error, "If shell charges are provided as guess, atomic dipoles (dpat) and quadrupoles (qmat) must also be provided")
```
pitsteinbach:
```suggestion
         call fatal_error(error, "If shell charges are provided as guess, atomic dipoles (dpat)"& 
                                             & //"and quadrupoles (qmat) must also be provided")
```

pitsteinbach (changes requested):
Currently you only provide testing for the C API can you also provide FORTRAN and python tests, please.

MauriceDHanisch:
@pitsteinbach I provided python and FORTRAN tests

include/tblite/result.h
```diff
@@ -262,4 +292,50 @@ tblite_save_result_wavefunction(tblite_error error,
...
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_set_result_shell_charges_and_moments_guess_spin(tblite_error error,
+                                                       tblite_result res,
+                                                       const double* qsh,
+                                                       const int nsh,
+                                                       const double* dpat,
+                                                       const int nat,
+                                                       const double* qmat,
+                                                       const int nat2,
+                                                       const int nspin);
```
pitsteinbach: the way I see it we can get away with using a single API call including spin, can you comment in which case two nat identifiers would be needed?

python/tblite/interface.py
```diff
@@ -274,6 +284,9 @@ def get(self, attribute: str):
          gradient               nat, 3                            Hartree/Bohr
          virial                 3, 3                              Hartree
          charges                nat                               e
+         shell-charges          nsh [2, nsh]                      e
+         atomic-dipoles         nat, 3 [2, nat, 3]                e·Bohr
+         atomic-quadrupoles     nat, 6 [2, nat, 6]                e·Bohr²
```
pitsteinbach, line 289: in the setter you reported the array dimensions using `(nsh, )` if we go with a general form then express it using spin

python/tblite/library.py
pitsteinbach, line 1: one thing that I am missing in terms of  functionality would be to restart a calculation from a results container directly, or is that possible?

src/tblite/api/result.f90
pitsteinbach, line 1: Here again it would be less code to generalise for any number of spins imo

test/api/main.c
pitsteinbach, line 1: You should also test for the error checks that you have created in the API, this partially leads to the low test coverage.

test/unit/meson.build
pitsteinbach, line 1: you also need to add the test file to the `CMakeLists.txt` file in this folder

pitsteinbach (changes requested):
So I would suggest to :
- always code for the general case with nspins the flattening can be done only in Python for example, in C this doesn't matter at all
- I would create a way in which a calculation can be started from a results container that is filled from a prior SCF
- to test that we are actually reusing the prior results we might want to track the number of iterations it took.
Thanks so far

include/tblite/result.h
MauriceDHanisch: The 2 nats is a remnant. I removed it! I also combined the 2 APIs

python/tblite/library.py
MauriceDHanisch: I am not sure if I understand. You could already restart from a results container previously

MauriceDHanisch:
The latest test that I added is currently failing but this is related to the mixer and not the implementation itself if I understand correctly. This is work in progress. I'll take another look later on but it would be nice to hear what you already think @pitsteinbach

PR #298, opened 2025-11-29
Fix spin-polarized restart issue

MauriceDHanisch:
This PR fixes the spin polarized restart issue. I have added a regression test to the python interface in `python/tblite/test_interface.py`.


test/test_restart_spin.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+import numpy as np
+from tblite.interface import Calculator, Result
+
+def test_spin_polarized_restart():
```
thfroitzheim: Please incorporate this test into the python tests (so likely into python/tblite/test_interface.py). Alternatively, you can start a new test file in the python directory just for the restart, but then we should make it more comprehensive.

thfroitzheim (changes requested):

MauriceDHanisch:
Sorry it took me so long to get back to this! I moved the test into the python interface and dropped the unused import. Does this look good to go? Let me know if you want me to squash the commits.

thfroitzheim (approved):

PR #299, opened 2025-11-30
Use compiliers/libraries not in $DEVTOOLSET_ROOT

### LINKED ISSUE #286: PyPI release might include broken extension module
Using tblite 0.5.0 from PyPI apparently can lead to

````
E   ImportError: tblite C extension unimportable, cannot use C-API
````

Thanks to Ryota @ryotatomioka for reporting.

### PR DESCRIPTION
**Pull request by yuzie007**
Closes #286

# Summary

This PR fixes the issue reported in #286.
I found the wheels of v0.5.0 for Linux are broken, likely due to the conflicts between `gcc-gfortran-8` used in LAPACK installed with `yum` and `gcc-gfortran-14` in `$DEVTOOLSET_ROOT` already installed in the `manylinux` image (AlmaLinux 8 in `manylinux_2_28`).
To avoid the conflicts, in this PR, I specify the ones consistent with LAPACK.
The tests are also added to ensure that the fixed wheels are really importable in Python.

# Details

The update of `cibuildwheel` to `3.3.0` (#297) ensures that it calls the `manylinux_2_28` image, which can install more recent LAPACK compared with `manylinux2014`.

The next commit 21c21abf6f55351960dfe961eaae9c2cf39e43e2 adds `CIBW_TEST_COMMAND` to test whether the created wheels are really importable.
This can avoid buggy wheels to be uploaded to PyPI in future.
At this point, as expected, the wheels show an error reported in #286. (Please refer to the results in my fork https://github.com/yuzie007/tblite/actions/runs/19797710364).

The last commit 87342ff6ed0db353a2e3ca6f8f1ddb1150e0c6c2 fixes the issue to use the compilers and the libraries compatible with LAPACK installed via `yum`.
After this fix, tests added above have passed (https://github.com/tblite/tblite/actions/runs/19799510554?pr=299).

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -136,6 +136,7 @@ jobs:
             LDFLAGS=-L/opt/homebrew/opt/openblas/lib
             CPPFLAGS=-I/opt/homebrew/opt/openblas/include
             MACOSX_DEPLOYMENT_TARGET=15.0
+          CIBW_TEST_COMMAND: python -c "import tblite.interface"
 
       # Upload the built wheels as artifacts
       - uses: actions/upload-artifact@v4
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -121,6 +121,20 @@ jobs:
           # Control verbosity of the 'pip wheel' output
           CIBW_BUILD_VERBOSITY: 1
           CIBW_BEFORE_ALL_LINUX: yum install -y lapack-devel
+          # Use compiliers/libraries not in $DEVTOOLSET_ROOT
+          # LAPACK installed by `yum install -y lapack-devel` in AlmaLinux 8
+          # in manylinux_2_28 may be compiled with gcc-gfortran 8.5.0, while
+          # AlmaLinux 8 also has GCC 14 from devtoolset-14 installed
+          # in $DEVTOOLSET_ROOT.
+          # To avoid conflicts, we specify LAPACK-compatible compilers/libraries.
+          # https://cibuildwheel.pypa.io/en/v3.3.0/options/#linux-image
+          # https://github.com/pypa/manylinux
+          CIBW_ENVIRONMENT_LINUX: >
+            CC=/usr/bin/gcc
+            CXX=/usr/bin/g++
+            FC=/usr/bin/gfortran
+            LD_LIBRARY_PATH=/usr/lib64
+            C_INCLUDE_PATH=/usr/include
           # Delete all other (=/= gcc-15) GCC versions to avoid conflicts in delocation of the wheel
           # MM: 'openblas' from brew requires gcc@15, using a different version for building leads to delocation errors
           # MM: gcc@14 needs macOS14

### CODE REVIEW

File: .github/workflows/wheel.yml [hunk 1]
```diff
@@ -121,6 +121,20 @@ jobs:
...
+          # in manylinux_2_28 may be compiled with gcc-gfortran 8.5.0, while
+          # AlmaLinux 8 also has GCC 14 from devtoolset-14 installed
+          # in $DEVTOOLSET_ROOT.
+          # To avoid conflicts, we specify LAPACK-compatible compilers/libraries.
+          # https://cibuildwheel.pypa.io/en/v3.3.0/options/#linux-image
+          # https://github.com/pypa/manylinux
+          CIBW_ENVIRONMENT_LINUX: >
+            CC=/usr/bin/gcc
+            CXX=/usr/bin/g++
+            FC=/usr/bin/gfortran
```
thfroitzheim, line 135: Let's maybe define the version we want to have the same as in MacOS (14 or 15 likely).

thfroitzheim (changes requested):

.github/workflows/wheel.yml
yuzie007: Hi @thfroitzheim thank you for the review.

There are several things I should clarify / re-emphasize.

First, I have confirmed that all modern gfortran 9, 10, 11, 12, 13, 14, 15 show the error essentially the same as #286. (I have observed in our Linux HPC clusters, [conda-forge feedstock](https://github.com/conda-forge/tblite-feedstock/pull/17), and even ReadTheDocs environment), as I reported my comment above and more specifically
```
  Traceback (most recent call last):
    File "/tmp/tmp.hUD1hKdDTn/venv/lib/python3.10/site-packages/tblite/library.py", line 30, in <module>
      from ._libtblite import ffi, lib
  ImportError: /tmp/tmp.hUD1hKdDTn/venv/lib/python3.10/site-packages/tblite/_libtblite.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZGVbN2v_exp
  
  The above exception was the direct cause of the following exception:
  
  Traceback (most recent call last):
    File "<string>", line 1, in <module>
    File "/tmp/tmp.hUD1hKdDTn/venv/lib/python3.10/site-packages/tblite/interface.py", line 26, in <module>
      from . import library
    File "/tmp/tmp.hUD1hKdDTn/venv/lib/python3.10/site-packages/tblite/library.py", line 32, in <module>
      raise ImportError(
  ImportError: tblite C extension unimportable, cannot use C-API
  Error: cibuildwheel: Command ['sh', '-c', 'python -c "import tblite.interface"'] failed with code 1. 
```
 This is the reason that we (so far) need to stick to an old gfortran (8.5.0).

Seeing the same issue in many environments, I think the error of `undefined symbol: _ZGVbN2v_exp` above (or `undefined symbol: _gfortran_os_error_at` in https://github.com/tblite/tblite/issues/286#issuecomment-3205880263) may be due to some recent changes in tblite between v0.4.0 and v0.5.0 (not in e.g. the gfortran side), i.e., some changes made tblite fragile with respect to the gfortran version.

I fully agree that it is ideal to use the same gfortran (as well as the same lapack) between Linux and MacOS, but in a short term the above issue may be difficult to solve, imagining the time limitation of the developers. My motivation is rather to offer a first-aid so that we can have v0.5.1 from PyPI rather soon (particularly because the present pip installation is broken for Linux as reported in #286).

The other aspect is the limitation of the `manylinux` image (AlmaLinux 8). This needs to be used to make the wheel compatible with many Linux environments. Likely due to this requirement, the `lapack` that can be installed with `yum` inside this image is compatible only with the old gfortran.

thfroitzheim: Ah, now I get the problem. Is it not enough to just exclude the build-in exponential function with something like 
```
CFLAGS="-fno-builtin-exp"
FFLAGS="-fno-builtin-exp"
```
Then we could keep using current compilers and wouldn't run into problems once the older compilers are no longer supported. If something like this doesn't work, we will have to go with the older gnu version.

yuzie007: Thank you @thfroitzheim for the feedback!

I can and will test e.g. `CFLAGS="-fno-builtin-exp"` and report you. Let me however already note that `_ZGVbN2v_exp` is just one example for `undefined symbol` I observed. Sometimes I saw `_ZGVbN2vv_pow` instead, and sometimes `_gfortran_os_error_at`, depending on the detailed compiler situation.

thfroitzheim: I see. Then let's check if it is possible with a reasonable number of exceptions (or maybe just removing all of them). Otherwise, we will use the old compiler

yuzie007: I found that `-fno-builtin` options are available only for C and not fortran do not solve the issue.

However, after a re-survey, I think I found a more reasonable solution (#300), so I would like to close this PR.

PR #297, opened 2025-11-26
Run `wheel.yml` for `macos-latest` again

yuzie007:
First of all, always thank you so much `tblite` developers and maintainers for your efforts!

# Summary

This PR suggests re-making the wheels for Apple Silicon macOS again.
With this, I would like to request to make them available again on PyPI, hopefully from the next release.

# Details

Wheels for Apple Silicon macOS were available in 0.4.0, and therefore pip-installation was easy.
```
% pip install tblite==0.4.0
Collecting tblite==0.4.0
  Downloading tblite-0.4.0-cp313-cp313-macosx_14_0_arm64.whl.metadata (5.8 kB)
Requirement already satisfied: cffi in ./miniforge3/envs/python-3.13/lib/python3.13/site-packages (from tblite==0.4.0) (1.17.1)
Requirement already satisfied: numpy in ./miniforge3/envs/python-3.13/lib/python3.13/site-packages (from tblite==0.4.0) (2.3.2)
Requirement already satisfied: pycparser in ./miniforge3/envs/python-3.13/lib/python3.13/site-packages (from cffi->tblite==0.4.0) (2.22)
Downloading tblite-0.4.0-cp313-cp313-macosx_14_0_arm64.whl (10.7 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 10.7/10.7 MB 6.1 MB/s eta 0:00:00
Installing collected packages: tblite
Successfully installed tblite-0.4.0
```
However, since [this commit](https://github.com/tblite/tblite/pull/271/commits/bf6ce48272f8f5557c728f1a28a905d78ea61758), wheels are not made for Apple Silicon macOS, and thus [PyPI does not have wheels for them](https://pypi.org/project/tblite/0.5.0/#files).
As a result, simple pip-installation fails as follows.
```
% pip install tblite==0.5.0
Collecting tblite==0.5.0
  Downloading tblite-0.5.0.tar.gz (861 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 862.0/862.0 kB 24.2 MB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [25 lines of output]
[...]

PR #301, opened 2025-12-05
Make Python wheels also for Intel MacOS

yuzie007:
This PR is a follow-up of #297.

This PR makes wheels also for Intel MacOS, under the same motivation as #297.

This also includes robustification of openblas-related-path specifications, in the same manner as [DFTD4](https://github.com/dftd4/dftd4/blob/v4.0.0/.github/workflows/wheel.yml), which is necessary as [the homebrew installation location is different between Intel and Apple-Silicon MacOS](https://docs.brew.sh/FAQ#why-should-i-install-homebrew-in-the-default-location).

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -76,10 +76,10 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        # macos-13
-        - macos-latest
-        # windows-latest
+          - ubuntu-latest
+          - macos-latest
+          - macos-15-intel
+          # windows-latest
         python: ['39', '310', '311', '312', '313', '314']
         include:
           - os: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


.github/workflows/wheel.yml
```diff
@@ -76,10 +76,10 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-latest
-        # macos-13
-        - macos-latest
-        # windows-latest
+          - ubuntu-latest
+          - macos-latest
+          - macos-15-intel
```
thfroitzheim, line 81: This is a good idea but if we are building wheels for Intel MacOS, we should incorporate it also in the build workflow to check that the tests actually pass.

yuzie007: I agree. I added the test for macos-15-intel in `build.yml` (so far with `gcc=14` and `meson`).

.github/workflows/wheel.yml
```diff
@@ -132,9 +132,9 @@ jobs:
           # Set macOS variables to find gfortran, lapack, and avoid testing against macOS earlier than 14 (see above)
           CIBW_ENVIRONMENT_MACOS: >
             CC=gcc-15 CXX=g++-15 FC=gfortran-15
-            PKG_CONFIG_PATH=/usr/local/opt/openblas/lib/pkgconfig
-            LDFLAGS=-L/opt/homebrew/opt/openblas/lib
-            CPPFLAGS=-I/opt/homebrew/opt/openblas/include
+            PKG_CONFIG_PATH="$(brew --prefix openblas)/lib/pkgconfig"
+            LDFLAGS="-L$(brew --prefix openblas)/lib"
+            CPPFLAGS="-I$(brew --prefix openblas)/include"
```
thfroitzheim, line 137: I just saw that we use the same absolute paths in the `build.yml`. Please replace those too.

thfroitzheim: see also line 96 in `build.yml`

thfroitzheim (approved):
LGTM

awvwgk (approved):

PR #305, opened 2025-12-07
Release version to 0.5.1

awvwgk:


yuzie007:
Thank you @awvwgk for putting things forward!

Should we maybe update the following as well?

https://github.com/tblite/tblite/blob/v0.5.0/python/ffi-builder.py#L41

Growl1234:
Hi
Would it also be possible to update dftd4 and s-dftd3 version in the wrapper file?

Issue #288, opened 2025-08-04
Test-job failure with tblite-0.5.0


sassy-crick:
I am trying to build `tblite-0.5.0`, i.e. the latest release, from source as part of a bigger project. I am using the following modules for that:
```
binutils/2.40-GCCcore-13.2.0 Meson/1.2.3-GCCcore-13.2.0 Ninja/1.11.1-GCCcore-13.2.0 FlexiBLAS/3.3.1-GCC-13.2.0
```
which are already installed on our cluster. As we are using `FlexiBLAS`, the build command is as follows: (*)
```
meson setup --prefix /tmp/tblite/0.5.0-gfbf-2023b  -Dlibdir=lib -Dlapack=custom -Dcustom_libraries='flexiblas'  -Db_ndebug=true ../build
```
The build is working well, however these test-jobs are failing consistently:
```
 81/165 tblite:unit / double-dictionary         FAIL            0.06s   exit status 1
 93/165 tblite:unit / npz                       FAIL            0.06s   exit status 1
104/165 tblite:unit / wavefunction-restart      FAIL            0.08s   exit status 1
165/165 tblite:api / C-API                      FAIL            1.33s   exit status 1
```
it does not matter if I am building on our AMD Rome nodes, or the Intel Icelake ones. I got similar problems on my laptop, where the test-jobs seem to fail more randomly. It is not an issue about timeout, as I have set the timeout quite generously:
```
meson test --num-processes 1 --print-errorlogs --no-rebuild -t 3
```
If it helps, I find that in the first failed job for example:
```
=================================== 81/165 ===================================
test:         tblite:unit / double-dictionary
start time:   13:59:31
duration:     0.06s
result:       exit status 1
command:      LD_LIBRARY_PATH=/rds/general/user/jsassman/home/easybuild/crest/source/tblite/build-icelake/:/sw-eb/software/binutils/2.40-GCCcore-13.2.0/lib:/sw-eb/software/Python/3.11.5-GCC
core-13.2.0/lib:/sw-eb/software/OpenSSL/1.1/lib:/sw-eb/software/libffi/3.4.4-GCCcore-13.2.0/lib64:/sw-eb/software/XZ/5.4.4-GCCcore-13.2.0/lib:/sw-eb/software/SQLite/3.43.1-GCCcore-13.2.0/li
[...]

PR #308, opened 2025-12-15
Fix the `C-API` issue for arrays of strings

yuzie007:
closes (maybe) #288

When I compile `tblite` with the `-fcheck=all` option on GitHub Actions CI ([here](https://github.com/yuzie007/tblite/actions/runs/20241791952/job/58111582732)), the `C-API` test fails for a job with the following message:
```
89/89 api - tblite:C-API FAIL 0.09s exit status 2 
...
Start test: table-builder 
stderr: 
At line 259 of file ../src/tblite/api/table.f90 
Fortran runtime error: Index '1' of dimension 2 of array 'carr' outside of expected range (1:0) 
```

I found this happens probably due to a typo in `src/tblite/api/table.f90:tblite_api_table:table_set_char_api`.
Specifically, this function is supposed to accept either a single string or a array of strings, and `n` (size of the array) should be `=0` and `>0` for the former and the latter cases, respectively.
Then, for the latter case, the `carr` pointer should have the shape corresponding to the array of strings.
However, the present implementation does not set the correct shape due to the typo of `max(n, 1)` as `min(n, 1)`.
The present PR fixes this.

After the fix above, the test in corresponding job in the CI suceeed ([here](https://github.com/yuzie007/tblite/actions/runs/20241849465/job/58111775749)).
I also confirmed the `C-API` failure on Windows on conda-forge feedstock is solved by the corresponding patch (please compare [this](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1418474&view=logs&j=fe7faf59-69bb-5487-b8fd-a35a7bda3a51&t=55731638-dd49-5144-2129-5101b9dd1960) with [the previous one](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1416358&view=logs&j=fe7faf59-69bb-5487-b8fd-a35a7bda3a51&t=55731638-dd49-5144-2129-5101b9dd1960)).
Hopefully, the present fix will also solve the `C-API` error in https://github.com/tblite/tblite/issues/288#issue-3289140647 (note that I believe the remaining `npz` related errors, including `double-dictionary` and `wavefunction-restart`, have already been solved in #307; please also refer to https://github.com/conda-forge/tblite-feedstock/pull/20#issuecomment-3640841958).

### DIFF
--- a/src/tblite/api/table.f90
+++ b/src/tblite/api/table.f90
@@ -244,7 +244,7 @@ subroutine table_set_char_api(verror, vtable, ckey, cval, n) &
 
    call c_f_pointer(vtable, table)
    call c_f_character(ckey, key)
-   call c_f_pointer(cval, carr, [strlen(cval)+1, min(n, 1)])
+   call c_f_pointer(cval, carr, [strlen(cval)+1, max(n, 1)])
 
    if (table%ptr%has_key(key)) call table%ptr%delete(key)
 


thfroitzheim (approved):
Good catch! But I am not sure that this will address #288 (maybe we should check that before closing the issue)

yuzie007:
Thank you @thfroitzheim for the feedback! I agree, it would be nice someone in #288 checks if the suggested fix solves their issue also on their environment.

PR #314, opened 2026-02-27
Make offline CMake installation valid 

Growl1234:
Offline CMake installation is not valid since multicharge and test-drive are not checked before checking dft-d4 and toml-f. This PR tries to fix this problem.


Growl1234:
Another question: Since "toml-f" needs "test-drive", it doesn't seem that the option `-DWITH_TESTS=OFF` is fully valid.

subprojects/test-drive.wrap
```diff
@@ -0,0 +1,4 @@
+[wrap-git]
+directory = test-drive
+url = https://github.com/fortran-lang/test-drive.git
+revision = v0.4.0
```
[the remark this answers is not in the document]
Growl1234: Other `subprojects/*.wrap` files still use `wrap-git`. This will temporarily remain unchanged.

Growl1234:
Gentle ping :)

P.S. I have to add `KMP_DUPLICATE_LIB_OK=TRUE` to pass all tests (tried [here](https://github.com/Growl1234/tblite/pull/1/checks) in my own fork; previously macOS wheel with older Python fails due to OpenMP conflict). I can also see another way which is to use acceleration instead of OpenBLAS, but I have concerns about it. Does anyone have better ways to resolve this problem?

> Another question: Since "toml-f" needs "test-drive", it doesn't seem that the option `-DWITH_TESTS=OFF` is fully valid.

And what do you think about this? This does not seem to be an obvious issue but maybe worth having a look.

config/cmake/Findtest-drive.cmake
thfroitzheim, line 1: I don't quite see the reason to introduce a test-drive dependency, since tblite itself doesn't use test-drive. It is actually a dependency of toml-f and comes in via the chain: mctc-lib => jonquil => toml-f. The resolution should happen somewhere downstream, as we otherwise introduce a possible source of version inconsistencies between tblite and its actual dependencies.

CMakeLists.txt
```diff
@@ -46,9 +46,15 @@ endif()
 if(NOT TARGET "mctc-lib::mctc-lib")
   find_package("mctc-lib" REQUIRED)
 endif()
-if(NOT TARGET "mstore::mstore" AND WITH_TESTS)
+if(NOT TARGET "mstore::mstore")
```
thfroitzheim, line 49: Why should we require the mstore dependency without tests?

CMakeLists.txt
```diff
@@ -46,9 +46,15 @@ endif()
...
   find_package("mstore" REQUIRED)
 endif()
+if(NOT TARGET "multicharge::multicharge")
+  find_package("multicharge" REQUIRED)
+endif()
+if(NOT TARGET "test-drive::test-drive" AND WITH_TESTS)
```
thfroitzheim, line 55: Here the same, we should only collect subprojects which are actual dependencies. For multicharge this is correct.

thfroitzheim (changes requested):

CMakeLists.txt
Growl1234: Because dftd4 requires mstore by default, and if here we don't remove `WITH_TESTS` tblite will still download mstore even if dftd4 has included it.

thfroitzheim: I don't see this. In tblite, we only need mstore in the tests. If dftd4 generally requires mstore, it should handle the dependency resolution. If dftd4 already got mstore, tblite won't get it a second time

config/cmake/Findtest-drive.cmake
Growl1234: That's a fair point. Perhaps this should only be implemented during the tarballing stage when cutting releases? We could set up the symlinks then, similar to what was done in [cp2k#4867](https://github.com/cp2k/cp2k/pull/4867/changes). The mstore and multicharge component might need a similar treatment. Or, directly move the folders to the subproject directories of the corresponding dependencies.

CMakeLists.txt
Growl1234: Sorry for confusion; to be clearer, this downloading behavior happens when `WITH_TESTS=OFF` is used (default is ON).

config/cmake/Findtest-drive.cmake
thfroitzheim: Yes, I agree. Since the main build system is meson and meson is also used to create the releases, cmake won't find the dependency in the right place. I think it is a good idea to add a distribution script to symlink mstore and maybe test-drive to the subprojects for distribution. I think multicharge should just be its own dependency since tblite always needs it.

Growl1234:
I'll submit an new PR for the alternative approach soon. The `find_package("multicharge" REQUIRED)` will be remained.

PR #328, opened 2026-05-06
Support optional smooth realspace cutoffs

### LINKED ISSUE #325: Support smooth cutoff
**Is your feature request related to a problem? Please describe.**
Due to the realspace cutoff a step function is introduced in the dispersion energy, which leads to jumps in the energy when atom distances move over the boundaries. This discontinuities can have impact on the gradient and stress tensor.

**Describe the solution you'd like**
Use a smooth cutoff which removes the step function on the cutoff boundary and keeps gradients and stress tensors well behaved.

**Additional context**
Developed for DFT-D4 and GFN2-xTB integration in CP2K by @DCM-Uni-Paderborn and @jpoto, see https://github.com/cp2k/cp2k/pull/5131

Requires support in https://github.com/dftd4/dftd4/issues/361 and https://github.com/dftd3/simple-dftd3/issues/182

Growl1234:
Attempts to resolve #325, along with updating required dftd4 and s-dftd3 version.

Fortunately, it seems some changes from the patch directly related to tblite is already in the development branch.


awvwgk (commented):
I would prefer to not read the smoothing width from an environment variable but rather set it directly when creating the calculator.

Also, the smoothing width needs to be passed to the D3 and D4 library to ensure the same smoothing is applied consistently.

awvwgk (approved):
Looks good to me now

PR #329, opened 2026-05-08
Refactor post-processing/xtbML

thfroitzheim:
This pull request refactors the post-processing/xtbML part. Both are needed for a deeper integration of `tblite` into `xtb` (e.g., for the property output)

Major changes:
- Refactor the full post-processing and xtbML implementation for clarity, reduced code duplication, and performance improvements.
- Separate immutable configuration (`post_processing_type`) from the output (`double_dictionary_type` in `result_type`) in the post-processing. The same post-processing instance can now be used for several singlepoints.
- Separate mutable intermediate data in the xtbML features in the `xtbml_cache`
- Pass `timer` from the singlepoint into the post-processing for memory safety with parallel singlepoint calculations (problem observed in the numerical hessian in xtb).
- Use the already defined `cache_list` to store the caches for post-processing.
- Fix bug in the xtbML orbital energy features (for the beta channel, chemical potential and gap were not reset during calculation)
- Separate the records used to parameterize the post-processing as a derived type (`post_processing_record`) to ensure the `post_processing_record_list` contains only post-processing related records
- Remove unused `dump_from_array`/`load_from_array` options in the post-processing related records
- Fix bug in unrestricted WBO (missing factor 2)
- Add a BLAS wrapper for `symm`

API breaking changes:
- The `add_post_processing` function now requires the `structure_type` as input

Changes after review: 
- I will move a few files to make the directory structure consistent. Right now, I have only changed the module naming. 

Future changes: 
- I will add the spherical-to-cartestian trafo in a separate commit to the basis set. 
- Maybe move the Molden-file output to tblite (something like the one I added to [xtb](https://github.com/thfroitzheim/xtb/blob/f61b9164db1e449cf128a26944df85f43b409957/src/printmold.F90#L346)). Since the Molden file still requires basis set information (later with g-xTB also cached intermediates), we might want to do this in tblite. 
- Add localized MOs as a further post-processing to support features such as docking in xtb or protonation/msreact in CREST (already working on this)
- Replace the xtbML energy features with a contribution resolved energy calculation. That would be a major change, since we would have to pass the result object to the `get_energy` functions. Not ideal, but we need something like this especially for the solvation free energy.


awvwgk:
Thanks for cleaning this code up. Regarding the transformation matrix in particular, I don't think this one belongs in the post processing output.

I would think it would be better to add a method to the basis set for constructing the transformation matrix if required. In xtb we can access the basis set from the tblite calculator if needed.

thfroitzheim:
Yes, as part of the basis set is also a possibility. With g-xTB it will be a bit more complicated, since the basis set requires some temporary data. Currently, I do this via a basis set cache, which would have to live on. But I think that this will be necessary in any case, since otherwise one cannot use the basis set after the singlepoint. I can reorganize this. Should we add the molden output also to tblite, or should we just provide the support for any host program?

Do you also have a different idea of where to put orbital localization? As part of the wavefunction?

awvwgk (approved):

PR #332, opened 2026-05-11
Allow setting smooth cutoffs when constructing calculator

awvwgk:
For allowing smooth cutoff configuration as implemented in CP2K for xTB via tblite. Defaults to smooth cutoff of 0.05 as used in CP2K.

- [x] expose smooth cutoff config in Fortran API
- [x] expose smooth cutoff config in C API
- [x] expose smooth cutoff config in Python API


awvwgk:
FYI @DCM-Uni-Paderborn @jpoto

PR #334, opened 2026-05-13
Expose MORFEUS-oriented descriptors via Python Result API

rlaplaza:
Add stable Result.get keys for covCN (CN_A), CM5 when stored in the result dictionary, a fractional-occupation-density proxy from MO occupations, and explicit accessors for Fukui / IP-EA / FOD that raise clear errors until those entries are populated. Extend the C API with tblite_get_result_covcn and tblite_get_result_cm5_charges backed by the Fortran result layer. Code may not be perfect and some other properties could and should be exposed too!

This lets MORFEUS and similar consumers rely on top-level keys instead of only post-processing-dict labels, while keeping existing keys unchanged. Ultimately, this is meant to reduce the near-infinite number of codebases that parse XTB output files and help migrate XTB-based workflows to tblite eventually.

### DIFF
--- a/include/tblite/result.h
+++ b/include/tblite/result.h
@@ -235,6 +235,26 @@ tblite_get_result_hamiltonian_matrix(tblite_error error,
                                      tblite_result res,
                                      double* hmat);
 
+/// Retrieve covalent coordination numbers from result container
+///
+/// @param error: Handle for error messages
+/// @param res: Result container
+/// @param covcn: Covalent coordination numbers, shape [nat]
+TBLITE_API_ENTRY void TBLITE_API_CALL
+tblite_get_result_covcn(tblite_error error,
+                        tblite_result res,
+                        double* covcn);
+
+/// Retrieve CM5 charges from result container
+///
+/// @param error: Handle for error messages
+/// @param res: Result container
+/// @param cm5: CM5 charges, shape [nat]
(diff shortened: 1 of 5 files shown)


thfroitzheim:
Thank you for the contribution, but I don't think that this can work in the current form.

First of all, things like CM5 charges, Fukui, or FOD are not implemented in tblite (also not in the post-processing). After all, tblite isn't a 1:1 replacement for xtb with every possible property.

But more importantly, the whole idea of the post-processing module is to avoid having explicit functions throughout the API for every single property. Instead, everything is collected into one consistent and exportable dictionary, which is then also translated to C and Python. So directly exposing properties to the API shouldn't be the goal. 

Having said this, we can think about adding some of the properties to the post-processing. CM5 charges might be an idea. IP-EA is more of a workflow for a host program with just three single points. Fukui and FOD are somewhere in between.

rlaplaza:
I misunderstood the goals! Buuuut certainly having charges and basically any other potentially relevant descriptor that one may use exposed in the post-processing-exportable dict would rock. In particular, for connecting with MORFEUS. Hit me up if i can help somehow, otherwise feel free to kill this PR.

thfroitzheim:
Right now, I plan to add orbital localization for compatibility in xtb, but nothing beyond this. But if there is a specific interest in some properties, please open an issue, and we can try. Of course, contributions are always welcome!

PR #339, opened 2026-05-21
Fix H0 strain derivative

thfroitzheim:
In the current version, the list of translations for the self-interaction (`iat==jat`)  includes both the forward and backward images (+T and -T). To avoid double-counting, we need a factor 1/4 instead of 1/2 (1/2 for the symmetrization and 1/2 for the double-counting). 

This wasn't clear since a factor of 1/2 was also missing in the numerical derivative test. 

Patch by @DCM-Uni-Paderborn


awvwgk (approved):

DCM-Uni-Paderborn:
Additional CP2K-side verification for this fix:

I reproduced the periodic virial discrepancy that motivated the check against CP2K's `REFERENCE_CLI` path. The issue is specific to the released `v0.6.0` H0 strain derivative and disappears with this PR on `main`.

For a periodic GFN2 Ar Gamma test, comparing CP2K's analytical `-pv_virial` to native `tblite run --grad` gives:

| native tblite | max virial diff vs CP2K | sum virial diff vs CP2K |
|---|---:|---:|
| v0.5.0 | `1.63e-4` | `5.98e-4` |
| v0.6.0 | `9.85e-3` | `2.95e-2` |
| main after #339 | `1.83e-7` | `7.30e-7` |

For a periodic GFN1 COBe Gamma test:

| native tblite | max virial diff vs CP2K | sum virial diff vs CP2K |
|---|---:|---:|
| v0.5.0 | `9.34e-7` | `1.76e-6` |
| v0.6.0 | `8.48e-2` | `2.49e-1` |
| main after #339 | `9.34e-7` | `1.76e-6` |

A finite-difference cell-strain check for the GFN2 case also follows the CP2K / v0.5.0 / post-#339 result rather than the `v0.6.0` analytical diagonal virial. So CP2K should not be adjusted to the `v0.6.0` CLI virial; the upstream H0 self-image strain prefactor fixed here is the right resolution.

PR #343, opened 2026-05-26
Fix multipole ewald + virials with updated scaling tests

lmseidler:
The old implementation had formal errors in the reciprocal space contributions:
1. it assumed rtrans included only +k and not -k
2. the charge-quadrupole didn't use the correct quadrupole prefactor
Accordingly, the reciprocal space contributions to the gradients and virials were also incorrect. This is corrected now.

Furthermore, the damping function introduces an algebraically decaying tail, which converges too slowly to use the same cutoff radius as the Ewald summation. I just set a high threshold for the real space cutoff which usually means that the sum converges to ~10e-7 accuracy. This introduces some overhead compared to previous calculations, however. The question is if we should just accept this or choose a shorter cutoff and accept higher errors (which would still require probably ~50 bohr cutoff radius to converge below mHartree accuracy).

What I also removed were the surface energy terms, which should be irrelevant considering that tblite only supports 3d pbc and conducting boundaries shouldn't introduce any errors.

Finally, I got rid of the usage of the minimum image convention/CCM since it doesn't make sense to use it like this here.

I enabled all unit tests and updated the scaling tests to calculate the energies dynamically, which is less prone to breaking on updates and targets the scaling more specifically. Also, the virials now converge to an acceptable accuracy.

### DIFF
--- a/src/tblite/coulomb/multipole.f90
+++ b/src/tblite/coulomb/multipole.f90
@@ -685,7 +685,9 @@ pure subroutine get_amat_sdq_dir_3d(rij, rr, kdmp3, kdmp5, alp, trans, &
 
       tmp = fdmp3 * g3 + e1
       amat_sd = amat_sd + vec * tmp
-      amat_dd(:, :) = amat_dd(:, :) + unity * tmp
+      amat_dd(1, 1) = amat_dd(1, 1) + tmp
+      amat_dd(2, 2) = amat_dd(2, 2) + tmp
+      amat_dd(3, 3) = amat_dd(3, 3) + tmp
       tmp = fdmp5 * g5 + e2
       amat_dd(:, 1) = amat_dd(:, 1) - vec * vec(1) * (3 * tmp)
       amat_dd(:, 2) = amat_dd(:, 2) - vec * vec(2) * (3 * tmp)
--- a/src/tblite/coulomb/multipole.f90
+++ b/src/tblite/coulomb/multipole.f90
@@ -430,7 +430,8 @@ subroutine get_dir_trans(lattice, alpha, conv, trans)
    !> Translation vectors
    real(wp), allocatable, intent(out) :: trans(:, :)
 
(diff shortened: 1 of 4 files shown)


src/tblite/coulomb/multipole.f90
```diff
@@ -430,7 +430,8 @@ subroutine get_dir_trans(lattice, alpha, conv, trans)
    !> Translation vectors
    real(wp), allocatable, intent(out) :: trans(:, :)
 
-   call get_lattice_points([.true.], lattice, get_dir_cutoff(alpha, conv), trans)
+   ! call get_lattice_points([.true.], lattice, get_dir_cutoff(alpha, conv), trans)
+   call get_lattice_points([.true.], lattice, 100.0_wp, trans)
```
awvwgk, line 434: It would be great if we could avoid this super large cutoff here. 100 Bohr is even larger than our default for dispersion (60 Bohr). If it is necessary, can we introduce a switching function which turns the damped multipoles at the cutoff value to a regular multipole Coulomb term? Similar as we do right now for switching the Klopman-Ohno electrostatic to a regular monopole Coulomb term.

lmseidler: not in a clean way. introducing a cutoff should be done with careful consideration, since for the 2nd order term this wrecked PESs for inorganic materials. it would be acceptable to choose a smaller cutoff, the large cutoff is most important for the tests where tight convergence is required. convergence to lower values could still be acceptable. i will check this

awvwgk: Do you think that using a taper function between damped multipoles and undamped multipoles would be an issue for large distances? We do this already for the isotropic Coulomb electrostatic as

image

This of course has its shortcomings as shown in *J. Chem. Theory Comput.* 2025, 21, 4, 1615–1625 [doi:10.1021/acs.jctc.4c01234](https://doi.org/10.1021/acs.jctc.4c01234).

PR #333, opened 2026-05-12
Molden file

thfroitzheim:
This PR adds a Molden file input and output. I try to follow the [official Molden specification](https://www.theochem.ru.nl/molden/molden_format.html). There are only two deviations necessary: 
- The Molden format lacks a specification for core electrons. Reducing the atomic charge breaks with the Molden format and is incompatible if the real nuclear charge is required. There are several different formats for added sections. I added the `[Core]` section (Molden2AIM and PySCF), the `[Nval]` section (Multiwfn and CP2K), and the `[Pseudo]` section (Molden and Orca).
- An additional `[Cell]` section consistent with CP2K and Multiwfn to specify periodic systems since Molden has no official specification for periodicity. 

The Molden format isn't fully identical to the xtb version, since xtb is missing a normalization for d functions (here all basis functions include full normalization in the coefficients) and has a wrong ordering of the Cartesian components. 

I aligned the whole implementation closely with the TREXIO reader/writer and added a similar test suite. 

I also fixed a small remaining bug in the TREXIO occupations for restricted calculations. 

closes #331

PR #344, opened 2026-05-29
Add missing parts to Trexio output

thfroitzheim:
I modified the TREXIO file handling, including the following points:
- Add the full basis set information, including contraction coefficients and exponents
- Add lattice vector output in the periodic case
- Core electrons are now part of the ECP block, and the total number of electrons includes the core electrons
- The order of the MO coefficients is now consistent with the TREXIO definition (so 0, 1, -1, ... l, -l and alphabetical in the Cartesian case).
- The TREXIO reader now populates the structure, basis, and wavefunction types directly since they are already available and ultimately decisive to check if the reader and writer are consistent. For this purpose, I added a restart test based on the read information. Maybe a TREXIO-based restart would generally be possible, but right now, TREXIO is still missing the electronic temperature. 

I will change the Molden implementation in #333 consistently with TREXIO in the IO part of tblite.

### DIFF
--- a/src/tblite/basis.f90
+++ b/src/tblite/basis.f90
@@ -24,11 +24,11 @@
 module tblite_basis
    use tblite_basis_ortho, only : orthogonalize
    use tblite_basis_slater, only : slater_to_gauss
-   use tblite_basis_type, only : basis_type, new_basis, get_cutoff, cgto_type
+   use tblite_basis_type, only : basis_type, new_basis, get_cutoff, new_cgto, cgto_type
    implicit none
    private
 
-   public :: basis_type, new_basis, get_cutoff, cgto_type
+   public :: basis_type, new_basis, get_cutoff, cgto_type, new_cgto
    public :: orthogonalize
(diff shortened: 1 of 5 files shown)


awvwgk (approved):

q-posev:
Hello @awvwgk and @thfroitzheim,

Happy to see that TREXIO found its way in `tblite`!

I am one of the core maintainers of TREXIO. Following up on your earlier comment:

> TREXIO is still missing the electronic temperature

I believe this should be a straightforward (single float number) addition to the TREXIO format. If this is useful for you - feel free to open a ticket in the TREXIO repository and we will integrate it in the next release. Maybe the `electron` group is the right place to host it?

cc @scemama

thfroitzheim:
Thank you for reaching out! 

I just opened an [issue](https://github.com/TREX-CoE/trexio/issues/263) in the TREXIO repo. Something like a single number in the electron group should be sufficient at least for tblite.

PR #346, opened 2026-06-12
CMake: Disable BUILD_TESTING for toml-f when WITH_TEST=OFF

Growl1234:
Package `toml-f` use a more general option name, `BUILD_TESTING`, rather than `WITH_TESTS`. Therefore, `WITH_TESTS=OFF` flag won't be passed to `toml-f`, and `toml-f` will keep building unittests. This PR fixes the behavior.


awvwgk (approved):
Thanks for sharing, looks good to me.

PR #350, opened 2026-06-18
Use smooth Wigner-Seitz truncation for periodic second-order electrostatics

lmseidler:
A fixed 10 Bohr cutoff for periodic second-order electrostatics causes discontinuities when periodic images enter or leave the cutoff sphere. It also truncates a significant part of the interaction energy relative to the periodic limit. This is visible as broken equations of state for ionic crystals:

NaCl GFN1 relative energy versus relative volume for fixed cutoffs and WSC truncation

## Changes

- Replace the fixed real-space cutoff with Wigner-Seitz-cell truncation for periodic second-order electrostatics.
- Use normalized compact C2 partition weights when several nearest periodic images compete, while preserving the corresponding lattice translation indices.
- Propagate the position and strain derivatives of those weights through the charge and multipole terms, yielding consistent analytic forces and stress at WSC transitions.
- Add finite-difference force and stress tests for structures that straddle a WSC transition, and update affected API, CLI, and Python references.

The corrected multipole Ewald implementation from #343 is retained; this PR applies smooth WSC image handling on top of the current implementation.

## Validation

- `cmake --build build --target tblite-tester -j 8`
- `ctest --test-dir build -R "^tblite/" --output-on-failure -j 4`: 35/35 tests passed
- Targeted `coulomb-charge` and `coulomb-multipole` suites, including WSC-threshold finite differences: 2/2 passed


awvwgk (dismissed):
Can we make `mic` an attribute of the coulomb type that is set on construction rather than an optional argument? In this case we can avoid changing the interface for the container and have a cleaner way to track of the property.

lmseidler:
Yes this makes sense, this is not at all final and still WIP. It is still not completely correct and needs some cleaning up.

lmseidler:
I'll implement some small fixes tomorrow to finish this but I can already say that this gives logarithmically divergent energies, as you would expect for a 1/r^3 tail. For ionic compounds, you can expect a variance of the order of 1e-2 eV per atom, which is not great but with the previous version the energies were wrong by a much larger magnitude anyways so I don't see this as a total loss. Relative energies would be inconsistent, accordingly. Despite this, the EOS shape is now qualitatively correct in all cases and equilibrium volumes hardly change, which is still more important in my opinion.

The correct thing to do would be using a proper Ewald scheme but I don't have time to do this at the moment since it is also not very trivial. The question is just whether we want to have the Ewald stubs like this in the official release or not. Perhaps we can postpone fixing this to a point where the Ewald summation is correctly implemlented.

DCM-Uni-Paderborn:
Added commit `80e1d14` to make Wigner-Seitz image transitions differentiable.

The update:

- preserves the actual translation indices after excluding the zero translation,
- replaces hard equal-image switching with normalized compact C2 weights,
- includes the position and strain derivatives of those weights in charge and multipole forces/stress,
- adds finite-difference force and stress tests directly at WSC image-transition thresholds.

Local validation:

```text
ctest --test-dir build -R 'coulomb-(charge|multipole)' --output-on-failure
2/2 tests passed
```

awvwgk (approved):

PR #354, opened 2026-06-27
Release version 0.7.0

awvwgk:
Thomas @thfroitzheim, Leopold @lmseidler, please let me know if you have any feature that should be part of tblite 0.7.0


thfroitzheim:
I think the fix in #350 would be good to have (not sure how far you are with this @lmseidler). Otherwise, nothing from my side.

PR #371, opened 2026-07-21
Switch integrals to CCA ordering

### LINKED ISSUE #352: Adopt CCA ordering for orbitals
It would be nice to consolidate the orbital ordering and switch to CCA which is used more or less consistently in libraries (libint, gauxc, libcint, etc.)

robin-dahl:
## Summary

This PR adds an optional [libcint](https://github.com/sunqm/libcint) Gaussian integral evaluator and introduces a common integral-handler infrastructure that allows tblite to select between its native implementation and libcint without embedding implementation-specific branches in the Hamiltonian routines.

It also changes tblite's internal Cartesian Gaussian ordering to the standard CCA convention (addressing #352) and updates the associated transformations and tests.

  ## Main changes

  ### Libcint integration

  - Adds Fortran bindings for the spherical libcint overlap, dipole, and quadrupole integrals and their nuclear derivatives.
  - Converts and caches tblite's contracted Gaussian basis in libcint's `atm`, `bas`, and `env` representation.
  - Handles the normalization, component ordering, origin conventions, and tensor transformations required to reproduce tblite's native integrals.
  - Compiles libcint with its `PYPZPX` flag to match tblite's real-spherical ordering.
  - Adds a `--libcint` command-line option.

  The dipole and quadrupole gradient routines currently require the `dp-qp-gradient` branch of [robin-dahl/libcint](https://github.com/robin-dahl/libcint), as these entry points are not available in the official libcint release. This dependency is intended to move to a fork owned by the tblite organization.

  ### Integral-handler infrastructure

  - Introduces a common abstract interface for Gaussian integral evaluation.
  - Provides native and libcint implementations through `native_integral_type` and `libcint_integral_type`.
  - Stores the selected evaluator in the xTB `calculator` and selects it through `set_integral_handler`.
  - Keeps the native evaluator as the default.
  - Keeps the Hamiltonian independent of the selected evaluator.

  ### Native integral organization

  - Moves the native implementation into `integral/native`.
  - Consolidates the former overlap, dipole, and multipole modules into a single native-integrals module (recommendations for a better module name are welcome).
  - Thereby removes duplicated integral routines.

  ### CCA Cartesian ordering (-> Closes #352)

  - Changes tblite's Cartesian Gaussian ordering to CCA: decreasing $\ell_x$, then decreasing $\ell_y$, with $\ell_z=\ell-\ell_x-\ell_y$.
  - Centralizes shell dimensions, offsets, and Cartesian exponent tables in `tblite_integral_shell`.
  - Updates the Cartesian/spherical and diatomic transformations and their consumers for the new ordering.

  ## Testing

  The libcint tests compare against tblite's native implementation for overlap, dipole, and quadrupole integrals and all their used derivatives. They cover every shell-pair combination from s through f.

  The tests also compare:

  - calls through the common integral-handler interface
  - complete GFN2-xTB energies obtained with both evaluators
  - Cartesian/spherical transformations using the new CCA ordering
  - builds with libcint disabled.

  ## Build support

  `Meson` and `CMake` support external libcint installations and a bundled fallback. The bundled dependency is handled correctly for shared, static, and Python wheel builds.

  `FPM` builds do **not** currently support libcint.


config/cmake/Findcint.cmake
```diff
@@ -0,0 +1,46 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with tblite.  If not, see <https://www.gnu.org/licenses/>.
+
+set(_lib "cint")
+set(_pkg "CINT")
+set(_url "https://github.com/robin-dahl/libcint.git")
```
awvwgk: Will this fork be upstreamed to https://github.com/sunqm/libcint?

test/unit/meson.build
```diff
@@ -71,7 +72,11 @@ test_srcs = files(
   'main.f90',
 )
 foreach t : tests
-  test_srcs += files('test_@0@.f90'.format(t.underscorify()))
+  if t == 'integral-libcint'
+    test_srcs += files('test_integral_libcint.F90')
+  else
```
awvwgk: Why is this needed?

config/meson.build
```diff
@@ -154,6 +155,45 @@ if trexio_dep.found()
   )
 endif
 
+libcint_found = false
+libcint_dep = declare_dependency()
+if not get_option('libcint').disabled()
+  libcint_dep = dependency('cint', required: false, method: 'pkg-config')
+  if not libcint_dep.found()
```
awvwgk: I think using a boolean instead of a feature makes more sense here.

awvwgk:
I would prefer to split this PR into two pieces, one patch for the CCA reordering and the refactoring of the native integral code and a second patch for the libcint integration. That would make the review process much more efficient. The CCA reordering and integral refactoring should be fast to merge and also be the largest diff due to the regression tests, with the libcint addition afterwards we can than check against the native implementation easily.

awvwgk (approved):
Looks good to me.

PR #385, opened 2026-07-29
Fix Fermi function

thfroitzheim:
This PR fixes a few issues with the Fermi function:

- Currently, we optimize for the integer HOMO level instead of the actual number of electrons, effectively rounding the number of electrons and preventing its use for fractional electron numbers. For double precision, the change to the actual `nel` value has no effect. 
- The Fermi energy contains an incorrect factor of 0.5, which we don't notice since `e_fermi` remains unused. 
- There are a few numbers lacking `_wp` declaration, leading to a float-to-double cast
- The threshold is relatively loose (around `1e-8`) even for double precision. Tightening the threshold actually changes results on a similar level as the cutoffs in most tests. Since there is no convergence issue here, I tightened the threshold for double precision. 

I also enter the Fermi function only if the temperature is actually non-zero.


awvwgk (approved):
Thanks, looks good to me.

PR #392, opened 2026-08-17
Enable building Python API with CMake

Growl1234:
See https://github.com/dftd4/dftd4/pull/393, https://github.com/dftd3/simple-dftd3/pull/230.


Growl1234:
Added `FindddX.cmake` to try resolving CI failure. Unfortunately it can't reuse the current `tblite-utils.cmake` because what ddX exported is `ddx`, not `ddx::ddx`.

2025-09-02 update: The CMake improvement has been merged upstream, but the above problem still exists as library name `ddX` is not consistent with target name `ddx`.

awvwgk (approved):

PR #396, opened 2026-09-17
Add orbital localization

thfroitzheim:
This PR adds Foster-Boys localized molecular orbitals and their centers to the post-processing methods. We need this for docking in xtb as well as for protonation and MSreact in CREST. Since this depends on details of the wavefunction and integral ordering, I think tblite is the right place for this. 

Localization uses a Jacobi solver with round-robin scheduling for parallelization. I also checked more advanced solvers based on block orbital rotation and L-BFGS or L-SR1, but the pairwise exact rotation with Jacobi is hard to beat (both for convergence speed and finding the highest maximum). The only thing that might improve here is a full second-order optimizer. 

Output works either through the post-processing dictionary or via Molden and TREXIO. 

Currently, LMOs are only supported for molecular systems since Foster-Boys isn't really useful for periodic systems. But the infrastructure makes it easy to add further localization schemes.


src/tblite/post_processing/localization.f90
```diff
@@ -0,0 +1,158 @@
...
+   type(orbital_localization), intent(out) :: self
+   !> Error handling
+   type(error_type), allocatable, intent(out) :: error
+   !> Integer identifier of the localization method to use
+   integer, intent(in) :: method_id
+
+   self%label = label
+   self%method_id = method_id
+
+   select case(self%method_id)
```
awvwgk: Can we use `method_id` directly here? Why do we need to store it in the type?

src/tblite/wavefunction/localization/type.f90
```diff
@@ -0,0 +1,333 @@
...
+         converged = .false.
+         return
+      end if
+
+      ! Refine the existing guess transformation via Jacobi sweeps
+      call self%optimizer%optimize(opmat, accuracy, trafo, spin_converged)
+      if (allocated(error)) then
+         converged = .false.
+         return
+      end if
```
awvwgk: Optimizer does not return an error, so there is no error to propagate

